PR #146, opened 2014-11-28
Sass 34/scss lint update

kaelig:
See #125 for details.


kaelig:
Will give this another go since gulp-ruby-sass was just updated to alpha2:
https://github.com/sindresorhus/gulp-ruby-sass/releases/tag/1.0.0-alpha.2

PR #147, opened 2014-11-28
Add support for Sass 3.4, and disable sourcemaps

kaelig:
This uses the latest available version of gulp-ruby-sass, so that we can test the Sass 3.4 integration easily.

As soon as it enters 1.0 release, we can put this into master.

(note: this is not a PR to go into master, but a separate testing branch: "sass-34")


kaelig:
Closing since gulp-ruby-sass v1 will change a lot of things and will require a larger rewrite than expected.

PR #166, opened 2014-12-18
More tests

AlbertoElias:
I've added tests for docs and demo, updated to latest master and other fixes. Test pass, but it seems to be slightly unstable. Would be nice to have it thoroughly reviewed.


lib/tasks/test.js
```diff
@@ -25,14 +25,14 @@ function silentCompilationTest(gulp, silent) {
 								reject(err.message);
 							})
 							.on('end', function() {
-								resolve();
+								resolve(true);
```
AlbertoElias: I forgot to remove it from there, thanks

AlbertoElias:
@matthew-andrews I fixed that
@kaelig Thanks! Though I think I messed it up again after merging master

Basically my only updates are 19c7c5b and 14121c2

matthew-andrews:
:+1:

PR #168, opened 2014-12-19
Demos: force edge mode in IE

kaelig:
Because we don't intend to support compat modes.


AlbertoElias:
I'll merge it, if the boilerplate hasn't been updated, @kaelig could you check it out?

kaelig:
@triblondon confirming the meta is in the suggested product boilerplate: http://origami.ft.com/docs/developer-guide/using-modules/#core-vs-enhanced-experience

PR #171, opened 2015-01-02
Adding sourcemaps option

aintgoin2goa:
I would like to have this options for Next as we need to have sourcemaps on production so that errors are tracable.

The sourcemaps config option can be a boolean
``` sourcemaps : true```
 or an object ie
    ```sourcemaps : {
        sass : false,
        js : true
    }```

As we can't yet have sass sourcemaps the 2nd method is pointless so I haven't documented it.

### DIFF
--- a/lib/tasks/build.js
+++ b/lib/tasks/build.js
@@ -30,11 +30,11 @@ module.exports.js = function(gulp, config) {
 			return newConfig;
 		}, {});
 
-		var sourcemaps = config.sourcemaps === true || config.sourcemaps && config.sourcemaps.js;
+		var useSourceMaps = config.sourcemaps === true || config.sourcemaps && config.sourcemaps.js;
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: lib/tasks/build.js [hunk 1]
```diff
@@ -30,9 +30,11 @@ module.exports.js = function(gulp, config) {
 			return newConfig;
 		}, {});
 
+		var useSourceMaps = config.sourcemaps === true || config.sourcemaps && config.sourcemaps.js;
```
kaelig: If I'm not mistaken, we don't need `config.sourcemaps.js` for now. Of course feel free to add it back when `config.sourcemaps` can also be an object.

aintgoin2goa: fair enough - was probably getting ahead of myself there.  Have removed it..

PR #175, opened 2015-01-07
Adding compress option

aintgoin2goa:
For next we want to have sourcemaps in production - this option will make this possible as then we can perform the minification ourselves.

You can specify `true` `false` or object ie: `{js:false,sass:true}`

If the compress option is not specified it falls back to the default behaviour.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ automatically re-build the module's demos and assets every time a file changes:
 All the tasks are built using [gulp](http://gulpjs.com/), and almost all of them return a stream. They are structured in 5 higher level tasks, and each one has one or more subtasks.
 
 	Usage: origami-build-tools <command> [<options>]
-	
+
 	Commands:
 	   install  Install system and local dependencies
 	   build    Build module in current directory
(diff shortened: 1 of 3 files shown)


AlbertoElias:
Is this an improved version of #171 ?

aintgoin2goa:
No, that option just generated sourcemaps - this one sets whether or not the output is compressed, overriding the env setting.

AlbertoElias:
but those two are what the development environment supports, why not just
use that?

*Alberto Elias*
Developer
FT Technology

+44 (0)74 75765609
Number One Southwark Bridge, London SE1 9HL


<https://www.facebook.com/financialtimes> <https://twitter.com/FT>
<http://www.linkedin.com/company/financial-times>
<https://plus.google.com/+FinancialTimes/posts>
<http://www.youtube.com/user/FinancialTimesVideos>


On 7 January 2015 at 17:09, Paul Wilson <notifications@github.com> wrote:

> No, that option just generated sourcemaps - this one sets whether or not
> the output is compressed, overriding the env setting.
>
> —
> <https://github.com/Financial-Times/origami-build-tools/pull/175#issuecomment-69054548>
> .
>

-- 

number 53723.

aintgoin2goa:
Yeah, that is a fair point.  It just seems odd to set env to 'development' for a production build, which is what I'd have to do.  Also, I think that it would be clearer to use explicit options for these behaviours.

I guess, it's just a matter of opinion, though - I can just use 'development' to achieve what I need.

AlbertoElias:
I think we should either have one or the other. @matthew-andrews @kaelig @triblondon What do you think?

kaelig:
I'm in favour of having the ability to toggle sourcemaps generation independently from the environment, with sensible defaults.

AlbertoElias:
Then should we look at deprecating the environment config option in favour of two separate compress and sourcemaps options which are basically what the environment option toggles

triblondon:
Fine with me.

PR #176, opened 2015-01-07
returning merged stream from build task

aintgoin2goa:
I came across an error whereby the build task reports itself as finished long before it actually has:
screen shot 2015-01-07 at 15 35 08

This makes it impossible to correctly chain other gulp tasks after obt.build.

Now the build method returns a merged stream of the 2 build tasks and this fixes the error:

screen shot 2015-01-07 at 15 35 55

screen shot 2015-01-07 at 15 36 07

### DIFF
--- a/lib/tasks/build.js
+++ b/lib/tasks/build.js
@@ -11,10 +11,13 @@ var gulpif = require('gulp-if');
 var autoprefixer = require('gulp-autoprefixer');
 var files = require('../helpers/files');
 var log = require('../helpers/log');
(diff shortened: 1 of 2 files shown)


AlbertoElias:
I believe this is possibly the case for all tasks. Could you check and, if so, update the other ones as well?

aintgoin2goa:
ok, sure

aintgoin2goa:
Crap, I must've merged and commented instead of just commenting - will undo

PR #181, opened 2015-01-08
Verbose

### LINKED ISSUE #160: Don't suppress successful output from bower install / npm install
Without this it makes it really hard to debug what has gotten installed in (say) CI logs.

matthew-andrews:
fixes #160

### DIFF
--- a/lib/tasks/install.js
+++ b/lib/tasks/install.js
@@ -82,7 +82,13 @@ function getCommandVersion(command, versionFlag) {
 	});
 }
 
-function cliSuccess() {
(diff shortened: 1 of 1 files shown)


lib/tasks/install.js
```diff
@@ -82,7 +82,13 @@ function getCommandVersion(command, versionFlag) {
 	});
 }
 
-function cliSuccess() {
+function cliSuccess(output) {
```
matthew-andrews: proabbly need to do `if (output && output.verbose){ `

PR #184, opened 2015-01-08
Add config option to make JSHint exit with code 1 when linting errors a...

robin-marr-ft:
...re found @AlbertoElias @matthew-andrews 

This was buggin me too so I've added a config option to allow this.

### DIFF
--- a/lib/tasks/verify.js
+++ b/lib/tasks/verify.js
@@ -4,6 +4,7 @@ var log = require('../helpers/log');
 var path = require('path');
 var fs = require('fs');
 var jshint = require('gulp-jshint');
+var gulpif = require('gulp-if');
 var scsslint = require('gulp-scss-lint');
 var lintspaces = require('gulp-lintspaces');
 var merge = require('merge-stream');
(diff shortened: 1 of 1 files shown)


robin-marr-ft:
when in dev it's nice to not have it exit so verification continues and you
see any lintspaces errors too, we could make it default true and leave the
option so people can chose.

On 9 January 2015 at 09:36, Alberto Elias <notifications@github.com> wrote:

> This is great, but I don't think it should be a config option, it should
> just be like that
>
> —
> <https://github.com/Financial-Times/origami-build-tools/pull/184#issuecomment-69312301>
> .
>

-- 

number 53723.

AlbertoElias:
We need to find a way to make sure it works similar to how it does if you run console commands (like we used to do) where it would run the different verify tasks, and if at least one failed, it would fail the whole thing but it would run all of them. scsslint and lintspaces also need to make sure they fail. Do you have any suggestions on how to do this?

robin-marr-ft:
I think this is possible but don't have time today, I'll take a look Monday
if that's ok?

On 9 January 2015 at 13:20, Alberto Elias <notifications@github.com> wrote:

> We need to find a way to make sure it works similar to how it does if you
> run console commands (like we used to do) where it would run the different
> verify tasks, and if at least one failed, it would fail the whole thing but
> it would run all of them. scsslint and lintspaces also need to make sure
> they fail. Do you have any suggestions on how to do this?
>
> —
> <https://github.com/Financial-Times/origami-build-tools/pull/184#issuecomment-69333032>
> .
>

-- 

number 53723.

AlbertoElias:
Sure, I need to get this sorted out for the testing branch as well and will look into it as soon as I finish some o-overlay issues

AlbertoElias:
This is fixed in the testing branch which will be pushed by Friday

PR #185, opened 2015-01-09
fixing error when there isn't both js and sass in a component

aintgoin2goa:


lib/tasks/build.js
```diff
@@ -14,10 +14,13 @@ var files = require('../helpers/files');
 var log = require('../helpers/log');
 
 module.exports = function(gulp, config) {
-	return merge(
-		module.exports.js(gulp, config),
-		module.exports.sass(gulp, config)
-	);
+	var jsStream = module.exports.js(gulp, config);
+	var sassStream = module.exports.sass(gulp, config);
+	if(jsStream && sassStream){
```
AlbertoElias: Would you mind adding a space surrounding the brackets for style consistency?

aintgoin2goa: sorry what do you mean?

AlbertoElias: `if (jsStream && sassStream) {` 

That

aintgoin2goa:
I'm guessing this is fixed in truck anyway - do we still need this pull request or shall I close it?

AlbertoElias:
Why do you think it's fixed in master?

aintgoin2goa:
Oh, it's not.  In that case, we really need to get this fix in as it is currently impossible to build a component that is only sass or only js

PR #195, opened 2015-01-23
Test all the things

AlbertoElias:
Test suite is ready. Seems to be quite stable.

@triblondon @matthew-andrews @samgiles Could you guys take a look and try doing clean installs of this branch, npm install, and then npm test a few times to see if it works fine please?


matthew-andrews:
it lands!

matthew-andrews:
checked in next-engels (an app) and next-header (a component) and works well!

lib/plugins/gulp-silent-sass.js
```diff
@@ -23,7 +27,7 @@ function gulpSilentSass(config) {
 	}
 
 	// creating a stream through which each file will pass
-	var stream = through.obj(function(file, enc, cb) {
+	stream = through.obj(function(file, enc, cb) {
```
samgiles: I'd be careful of using module scope to share objects like this, particularly when there is a chance that 'stream' could be undefined.  In this narrow example it might be better to create stream in gulpSilentSass and pass it as a parameter?

lib/tasks/build.js
```diff
@@ -109,7 +111,9 @@ module.exports.sass = function(gulp, config) {
 			.pipe(gulpif(config.env === 'production', csso()))
 			.pipe(rename(dest))
 			.pipe(gulp.dest(destFolder))
-			.on('error', function(err) { console.log(err.message);});
+			.on('error', function(err) { throw Error(err.message); });
```
samgiles: Why not just 'throw err'?

samgiles:
Looks good!  Nice work.    I've tested with `npm test` a few times and it was successful.   Feel free to merge when ready.

lib/tasks/build.js
AlbertoElias: Good point!

lib/plugins/gulp-silent-sass.js
AlbertoElias: I think you're right. I'll change it

AlbertoElias:
@triblondon I've improved the readme slightly. Good to merge and release?

triblondon:
Merging this is def not blocked on docs, so merge away. Will take another look at docs later.=

PR #197, opened 2015-01-27
[Sass 3.4] Disable sourcemaps using an undocumented technique

kaelig:
**Only works with Sass 3.4.**

When running with Sass 3.3, it returns:

```
[16:15:23] gulp-ruby-sass: stderr: OptionParser::NeedlessArgument: needless argument: --sourcemap=none
  Use --trace for backtrace.
```

### DIFF
--- a/lib/tasks/build.js
+++ b/lib/tasks/build.js
@@ -81,7 +81,9 @@ module.exports.sass = function(gulp, config) {
 		config.env = config.env || 'development';
 
 		var sassConfig = {
-			loadPath: ['.', 'bower_components']
+			loadPath: ['.', 'bower_components'],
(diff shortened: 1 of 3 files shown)


kaelig:
I see we're using `getCommandVersion('sass', '--version')` in `getInstalledSassGemVersion`.

It'd be great to make that reusable inside build.js and demo.js so that we can do something like:

```js
var sassVersion = getInstalledSassGemVersion();
if (semver.lt(sassVersion, '3.4')) {
  sassConfig["sourcemap=none"] = true;
}
```

And then this branch would work in both Sass 3.3 and 3.4.

@AlbertoElias what do you think?

kaelig:
Closing to reopen and merge into a sass-34 branch

PR #199, opened 2015-01-27
[Sass 3.4] Disable sourcemaps

kaelig:
Works only with Sass 3.4

@alberto:

I see we're using `getCommandVersion('sass', '--version')` in `getInstalledSassGemVersion`.

It'd be great to make that reusable inside build.js and demo.js so that we can do something like:

```js
var sassVersion = getInstalledSassGemVersion();
if (semver.lt(sassVersion, '3.4')) {
  sassConfig["sourcemap=none"] = true;
}
```

And then this branch would work in both Sass 3.3 and 3.4.

### DIFF
--- a/lib/tasks/build.js
+++ b/lib/tasks/build.js
@@ -81,7 +81,9 @@ module.exports.sass = function(gulp, config) {
 		config.env = config.env || 'development';
 
 		var sassConfig = {
-			loadPath: ['.', 'bower_components']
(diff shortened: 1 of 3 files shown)


kaelig:
@wheresrhys fixing the version detection problem would allow for Sass 3.3 and 3.4 support. Do you want to have a go (see comment up)?

wheresrhys:
@kaelig I'll have a go hopefully soon

PR #204, opened 2015-01-31
Adds command for generating bowerrc

wheresrhys:
Quicker than pointing people to registry to copy and return to terminal/finder.

I kinda want to put in a warning that it'll overwrite any existing bowerrc, but I think that's overcomplicating things for most people. @AlbertoElias


AlbertoElias:
I think it's important to be as explicit as possible with the least possible words. I would also mention that it does it in the home directory just in case :P

matthew-andrews:
i'm not against this but considering how often people need to do this (not often) tis an odd thing to optimise.

AlbertoElias:
I'm with @matthew-andrews on this. I'm going to add this to the trouble shooting guide though

wheresrhys:
:+1:

PR #193, opened 2015-01-21
Use origami.json browserFeatures property to use the polyfill service.

matthew-andrews:
This seemed like a good idea.

PR #205, opened 2015-01-31
6to5

matthew-andrews:
Plus a bunch of test race conditions

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "node": ">= 0.10.0"
   },
   "scripts": {
-    "test": "mocha --recursive -t 50000 *.test.js"
+    "test": "mocha -t 50000 `find . -name '*.test.js' -path './test/*'`"
   },
   "dependencies": {
     "6to5ify": "^4.0.0",
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "colors": "^0.6.2",
     "configstore": "^0.3.1",
     "debowerify": "Financial-Times/debowerify",
+    "denodeify": "^1.2.0",
     "es6-promise": "^1.0.0",
     "findit": "^1.1.1",
     "gulp": "^3.8.7",
--- a/package.json
+++ b/package.json
@@ -20,7 +20,6 @@
     "colors": "^0.6.2",
     "configstore": "^0.3.1",
     "debowerify": "Financial-Times/debowerify",
-    "denodeify": "^1.2.0",
     "es6-promise": "^1.0.0",
     "findit": "^1.1.1",
     "gulp": "^3.8.7",
@@ -51,6 +50,7 @@
     "which": "^1.0.5"
   },
   "devDependencies": {
+    "denodeify": "^1.2.0",
     "expect.js": "^0.3.1",
     "fs-extra": "^0.15.0",
     "mocha": "^2.0.1",
--- a/package.json
+++ b/package.json
@@ -11,9 +11,10 @@
     "node": ">= 0.10.0"
   },
   "scripts": {
-    "test": "node_modules/mocha/bin/mocha --recursive -t 50000"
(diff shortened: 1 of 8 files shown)


matthew-andrews:
oh actually this isn't really working

package.json
```diff
@@ -11,14 +11,16 @@
     "node": ">= 0.10.0"
   },
   "scripts": {
-    "test": "node_modules/mocha/bin/mocha --recursive -t 50000"
```
matthew-andrews: this was also matching the fixtures

test/tasks/demo.test.js
```diff
@@ -126,36 +129,34 @@ describe('Demo task', function() {
...
-		it('should build demo html', function(done) {
+		it('should build demo html', function() {
 			fs.writeFileSync('demos/src/test1.mustache', '<div>test1</div>', 'utf8');
 			fs.writeFileSync('demos/src/test2.mustache', '<div>test2</div>', 'utf8');
-			demo(gulp)
+			return demo(gulp)
 				.then(function() {
 					var test1 = fs.readFileSync('demos/test1.html', 'utf8');
-					var test2 = fs.readFileSync('demos/test1.html', 'utf8');
+					var test2 = fs.readFileSync('demos/test2.html', 'utf8');
```
matthew-andrews: I actually broke this in this PR:- https://github.com/Financial-Times/origami-build-tools/pull/193/files#diff-6a5945ca98ba77b07f1c75a9d4888aebR135

Which shows even though Travis has been telling us things are good, the tests haven't actually been passing for a few days…

I don't think the `npmTest` task works properly :-) @AlbertoElias

matthew-andrews:
Hmm.  Because of the build service we might not want to make ES6 the default behaviour just yet.

Maybe I can change it to be opt-in so that products can use it?

test/tasks/build.test.js
```diff
@@ -164,10 +161,7 @@ describe('Build task', function() {
 		});
 
 		afterEach(function() {
-			if (fs.existsSync('build/main.css')) {
-				fs.unlink('build/main.css');
-				fs.rmdir('build');
-			}
+			return exec('rm -rf build');
```
AlbertoElias: rimraf is a dependency for these kind of things

test/tasks/demo.test.js
```diff
@@ -39,27 +40,29 @@ describe('Demo task', function() {
 	});
 
 	describe('Build demos', function() {
-		it('should fail if there is not a bower.json file', function(done) {
-			demo(gulp)
-				.catch(function(err) {
+		it('should fail if there is not a bower.json file', function() {
+			return demo(gulp)
```
AlbertoElias: Why add a return statement?

matthew-andrews: mocha supports promises, this means that if the test fails it fails immediately rather than just timing out.

test/tasks/build.test.js
matthew-andrews: ?  this is just bash

AlbertoElias: I don't mind having denodeify or rimraf as a dependency, but I do think it makes sense to use the same one every time.

matthew-andrews: rimraf isn't a dependency?

https://github.com/Financial-Times/origami-build-tools/blob/master/package.json#L16-56

test/tasks/demo.test.js
matthew-andrews: also it's arguably simpler

test/tasks/build.test.js
AlbertoElias: Oh, right, I changed it to fs-extra which uses rimraf underneath. (`fs.rmdir`). That needs to stay for other uses, so I think it's fine to have denodeify as well.

PR #209, opened 2015-02-06
checking for semver compatibility rather than less than-ness

wheresrhys:
@AlbertoElias @kaelig


AlbertoElias:
Doesn't seem to work, didn't pass Travis

wheresrhys:
@alberto fixed it, though maybe in a slightly inelegant way

AlbertoElias:
Why not just change the versions on the versions object?

triblondon:
I would guess Rhys has a later version already installed and wants to avoid obt downgrading it.

wheresrhys:
And the opposite too - I want to avoid upgrading every time there is a patch release. More than likely these include fixes for edge cases that won't affect me. If there is an important fix that everyone should be forced to pick up obt should update the semvers here.

wheresrhys:
Incidentally if either of you know of a way of installing a gem using a semver range I'll happily substitute that method in here. A cursory search didn't turn anything up, but I'd be surprised if there isn't a proper semver implementation built into gem

AlbertoElias:
I'll discuss this with Rhys on Thursday

wheresrhys:
nudge @AlbertoElias @triblondon @kaelig Who's in OSB next so we can discuss and get it merged?

AlbertoElias:
@wheresrhys Oh yes, let's do it this Friday

wheresrhys:
:+1::+1:


On 24 March 2015 at 11:39, Alberto Elias <notifications@github.com> wrote:

> Merged #209
> <https://github.com/Financial-Times/origami-build-tools/pull/209>.
>
> —
> <https://github.com/Financial-Times/origami-build-tools/pull/209#event-263399119>
> .
>

-- 

number 53723.

AlbertoElias:
I'm going ot change it though so the minimum Sass is 3.4, and scss-lint is 0.28, to avoid issues, I think we should set 3.4 as the baseline

wheresrhys:
Ah yes, sorry, I missed that change. origami must've switched to 3.4 since my original PR

PR #210, opened 2015-02-06
Sass 3.4

kaelig:
Feel free to use the build tools with this branch. I've been using it for two weeks and it worked well so far.

The main benefit is that we can now use scss-lint's `scss-lint:disable` flag, such as:

```scss
	// scss-lint:disable SelectorDepth
	[data-o-hierarchical-nav-level='1'] {
		> li > a {
			line-height: 25px;
			max-height: 25px;
		}

		// Set styling for selected and hovered a tags
		[aria-selected="true"] a,
		[aria-controls][aria-expanded="true"] a,
		#{$o-hoverable-if-hover-enabled} & > li > a:hover {
			background-color: $_o-header-secondary-selected;
			color: oColorsGetColorFor(link-title, text);
			opacity: 1;
		}
	}
	// scss-lint:enable SelectorDepth
```

https://github.com/Financial-Times/o-ft-header/blob/3.0.0-beta.3/src/scss/nav/theme-secondary.scss#L20

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,8 +73,8 @@ Runs:
 
 The versions that are installed and supported are:
 
-* Sass: '3.3.14' _(Sass 3.4.x is currently not supported and you may not get the desired result)_
-* scss-lint: '0.27.0'
+* Sass: '3.4.x'
+* scss-lint: '0.33.0'
 * JSHint: '2.5.6'
 * Bower: '1.3.12'
 
--- a/config/scss-lint.yml
+++ b/config/scss-lint.yml
@@ -1,36 +1,72 @@
 linters:
 
-  # ENABLED:
+  # Documentation:
+  # https://github.com/causes/scss-lint/blob/master/lib/scss_lint/linter/README.md
+
+  # "value !important;" not "value!important ;"
+  BangFormat:
+    enabled: true
+    space_before_bang: true
+    space_after_bang: false
+    severity: error
 
   # "border: 0" not "border: none"
   BorderZero:
     enabled: true
+    severity: error
 
-  # No capitals in selectors
-  CapitalizationInSelector:
-    enabled: true
+  # Allow colors to be used outside of variables
+  ColorVariable:
+    enabled: false
+
+  # "#0f0" not "green"
+  # (AB: Causes problems with o-colors < 3, because o-colors uses
+  # color names that are also valid CSS color strings)
+  ColorKeyword:
+    enabled: false
+
+  # // not /**/
+  Comment:
+    enabled: false
 
   # Disallow "@debug"
   DebugStatement:
     enabled: true
+    severity: error
 
   # @extend must come first, then properties, then nested rulesets
   DeclarationOrder:
     enabled: true
 
-  # Disallow use of the same property more than once in the same ruleset
+  # Warn when the same property appears more than once in the same ruleset
+  # This is not a deal breaker, as authors may repeat properties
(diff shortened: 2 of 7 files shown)


AlbertoElias:
Couple questions. Will this work with an older scss-lint or does everyone need to update it? And is it basically the sourcemap hack? And it isn't passing Travis, npm test seems to fail

kaelig:
- This won't work with SCSS-Lint < 0.33
- It's the sourcemap hack, you are correct
- Not sure what's happening with travis… cf below, not very helpful


```
Error: expected 'object' to equal 'undefined'
      at Assertion.assert (/home/travis/build/Financial-Times/origami-build-tools/node_modules/expect.js/index.js:96:13)
```

AlbertoElias:
@kaelig I restarted Travis and it passed, weird.. We need a way to alert users that they need to run obt install after updating obt. Actually, this might require a major update of obt. I think Next use obt for their builds, and this update might break their builds.

@matthew-andrews Any insight on this?

matthew-andrews:
why might it break our builds — we always run `origami-build-tools install` straight after doing `npm install origami-build-tools`

AlbertoElias:
Brilliant! Then it might only break if you don't pass the new verify or/and if you're not compatible with Sass 3.4. Can we merge then?

matthew-andrews:
let me try it out on an app first.  can i have an hour or two?

AlbertoElias:
Of course, this isn't urgent

matthew-andrews:
about to run

```
nvm install 0.10.33
nvm use 0.10.33
gem install sass --version 3.4
npm install Financial-Times/origami-build-tools#sass-34
origami-build-tools --version
make install
```

on next-grumman

matthew-andrews:
is this branch backwards compatible with sass 3.3?

kaelig:
You might want to also install the latest scss-lint.

This branch could be made compatible with Sass 3.3.

matthew-andrews:
We actually rely on origami build tools to install scss lint so that happens automatically…  I can't remember why we decided to install sass not from OBT.

matthew-andrews:
works fine with the correct sass.. I'm going to try with:-

nvm install 0.10.33
nvm use 0.10.33
gem install sass --version 3.3
npm install Financial-Times/origami-build-tools#sass-34
origami-build-tools --version
make install

I think this as long as this either works or fails hard—ie. fails the build and prevents a deploy that's fine.

matthew-andrews:
i think we're good to go!

AlbertoElias:
@matthew-andrews What was the result of that last test?

matthew-andrews:
it build correctly without errors.

kaelig:
Before the release we should send a note to everybody saying they should run `gem install sass:3.4.11 scss-lint:0.33.0` if they want to run the latest version of OBT

AlbertoElias:
Oops, too late. But we should let people know. I'll send an email to
frontend devs

-- 

number 53723.

quarterto:
Just installed OBT 2.9. Even with Sass 3.4.11 and scss-lint 0.33 I'm still getting the source map error. Am I missing something? Running it as `make build` in next-engels.

AlbertoElias:
Can you run sass -v to check which version of Sass is being used?

quarterto:
```
❯ sass -v
Sass 3.4.11 (Selective Steve)
```

kaelig:
For reference: @quarterto's build tools magically started working again.

PR #214, opened 2015-02-10
Added custom fail reporter so build does not fail with warnings

### LINKED ISSUE #213: Warnings are being promoted to errors
We're getting closer :smile: 

screen shot 2015-02-10 at 15 33 37

Possibly:-
- Warnings (e.g. above) shouldn't break builds

Nicer solution?
- Maybe a `--strict` mode where all warnings become errors

AlbertoElias:
Fixes #213

### DIFF
--- a/lib/tasks/verify.js
+++ b/lib/tasks/verify.js
@@ -6,6 +6,8 @@ var fs = require('fs');
 var jshint = require('gulp-jshint');
 var scsslint = require('gulp-scss-lint');
 var lintspaces = require('gulp-lintspaces');
+var through = require('through2');
+var gutil = require('gulp-util');
 var merge = require('merge-stream');
 var cwd = process.cwd();
 var excludePath = path.join(cwd, '/.gitignore');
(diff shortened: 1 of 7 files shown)


lib/tasks/verify.js
```diff
@@ -60,17 +62,37 @@ function pathsToGlob(paths) {
...
 
 /**
+ * Custom fail reporter that doesn't fail on Scss Lint warnings
+ */
+function failReporter() {
+	var stream = through.obj(function(file, enc, cb) {
+		var error;
+		if (file.scsslint.errors > 0) {
+			error = new gutil.PluginError('gulp-scss-lint', {
+				message: 'ScssLint failed for: ' + file.relative,
```
kaelig: SCSS-Lint

lib/tasks/verify.js
```diff
@@ -60,17 +62,37 @@ function pathsToGlob(paths) {
 }
 
 /**
+ * Custom fail reporter that doesn't fail on Scss Lint warnings
```
kaelig: SCSS-Lint

kaelig:
GG :+1:

lib/tasks/verify.js
AlbertoElias: Actually, that line of code is copied from the fail reporter gulp-scsslint offers. But I will update anyway

PR #215, opened 2015-02-10
Force lowercase on first character of mixin, variable and function names

kaelig:


AlbertoElias:
:+1: Merge it as soon as Travis finishes :)

PR #220, opened 2015-02-23
Update SCSS-Lint from 0.33.0 to 0.34.0

kaelig:
- Add an option to exclude files from singular tests
- Upgrade scss-lint
- Update tests and the gulp module


kaelig:
Note that I simplified the custom scss-lint.yml config file to make it clearer what this is actually testing and easier to maintain.

AlbertoElias:
Thanks!

README.md
```diff
@@ -144,8 +144,10 @@ Runs:
 
 * __scssLint(gulp, config)__ Config accepts:
 	- scssLintPath: `String` Path to your custom 'scss-lint.yml' config file. (Default: 'origami-build-tools/config/scss-lint.yml') _This may be set for product development, but developers of Origami-compliant components are required to accept the default_
+	- excludeFiles `Array` e.g. `['!**/demo.scss']`
 * __jsHint(gulp, config)__ Config accepts:
 	- jsHintPath: `String` Path to your custom jsHint config file. (Default: 'origami-build-tools/config/jshint.json' _This may be set for product development, but developers of Origami-compliant components are required to accept the default_
+	- excludeFiles `Array` e.g. `['!**/demo.js']`
```
kaelig: @AlbertoElias I did not write tests for these. Should we have any?

AlbertoElias: @kaelig I'd say they're not essential, as they're mainly a gulp feature, but it's also quite easy to test, so whatever you prefer

kaelig: Right. I guess it's simple enough (a concat from a global array…) for not requiring tests.

PR #221, opened 2015-02-24
Bump scss-lint version up

kaelig:


kaelig:
How do I also merged it in the node 0.10 branch?

PR #228, opened 2015-03-10
Added notes for installing ruby on mac

adambraimbridge:


kaelig:
@adambraimbridge good instructions thanks.

Have you had an issue with installing Ruby yourself or another person did?

Should we output these instructions in the terminal when the failure happens instead?

AlbertoElias:
We're going to set up #229 and this will be a great addition to a new troubleshooting guide

AlbertoElias:
@adambraimbridge I added it in #238, so I'll close this pr. Thanks a lot for the guide! :)

PR #231, opened 2015-03-18
#perfmatters

samgiles:
Introduce a TaskLoader for loading tasks.
Introduce a new log level 'debug' (which is defaulted to 'off').  To turn it on run `DEBUG=true obt task`.

The main reason to introduce the DEBUG=true log level is to surface performance issues and help spot issues, if any are low hanging fruit, and easy to fix, we can fix them in the right place.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,5 +205,16 @@ npm install --save-dev gulp
 npm install -g gulp
 ```
 
+
+The `origami-build-tools` module also defines some helper methods to verify and list the available tasks:
+
+`obt.isValid(taskName)` will return a boolean value indicating whether the
+name of the given task is valid.
+
+`obt.list()` will return a list of valid task names.
+
+`obt.loadAll()` will return an object with all of the available tasks loaded onto it.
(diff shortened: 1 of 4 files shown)


samgiles:
Need to fix the tests f you agree this is a good idea.

AlbertoElias:
I really like it, but it adds the problem that this is a breaking change and it's slightly harder to use outside of the cli. Before you would do:

```js
var obt = require('origami-build-tools');
obt.build.js();
```

not it has to be:

```js
var obt = require('origami-build-tools');
obt.load('build').js();
```

This will affect gulpfiles

AlbertoElias:
Also, have you calculated the speed improvements before this pr? I'm curious

samgiles:
There are no speed improvements for running `build` as that's the heaviest one, you can see a huge difference when running `verify` or `--version` (`--help` has to load everything to see the descriptions so you can use that as a comparison).


I've updated to be backwards compatible.  Not sure why the build is still failing?

PR #233, opened 2015-03-25
Browser tests

AlbertoElias:
Now runs nightwatch on our selenium grid agains the demos running in a heroku instance.

@spriggsha @matthew-andrews @samgiles Could you take a look?


AlbertoElias:
Sorry about the git history, I messed it up quite a bit, tried to clean it up a little, but still not very nice

README.md
```diff
@@ -127,6 +126,7 @@ Config:
 * local: `Boolean` Build local HTML, CSS and JS files, in addition to demo HTML for the build service. Also runs a local server to help you test your demos.
```
samgiles: Should probably remove the mention of running the server now that runServer is there.  Perhaps `runServer` could imply `local`?

lib/tasks/test.js
```diff
@@ -60,17 +82,115 @@ module.exports.nonSilentCompilation = function(gulp) {
...
-	var packageJson = files.getPackageJson();
-	if (packageJson && packageJson.scripts && packageJson.scripts.test) {
-		commandLine.run('npm', ['test'])
-			.then(function(output) {
-				log.primary('Running "npm test"...');
-				log.secondary(output.stdout);
-			}, function(output) {
-				log.primary('Running "npm test"...');
-				log.secondaryError(output.stderr);
+	return new Promise(function() {
```
samgiles: I'd suggest resolving or rejecting this promise.  Seems weird to return a Promise that will never resolve.

lib/tasks/test.js
```diff
@@ -60,17 +82,115 @@ module.exports.nonSilentCompilation = function(gulp) {
...
+		demo(gulp, { local: true })
+	])
+	.then(function(results) {
+		token = results[0].stdout ? results[0].stdout.trim() : results[0].trim();
+		commit = results[1].stdout.trim();
+		appName = appName + commit.substr(0, 5);
+		return haikro.create({
+			app: appName,
+			region: 'eu',
+			// organization: 'financial-times',
```
samgiles: Why's this commented out?

lib/tasks/test.js
```diff
@@ -60,17 +82,115 @@ module.exports.nonSilentCompilation = function(gulp) {
...
+		return module.exports._runBrowserTest(config);
+	})
+	.then(function(output) {
+		cleanupTest(output);
+	})
+	.catch(function(e) {
+		if (e.stdout) {
+			cleanupTest(e)
+				.then(function() {
+					process.exit(1);
```
samgiles: :( `process.exit` makes me sad.  We should think about how we can have a cleaner interface in the future.

lib/tasks/test.js
AlbertoElias: I can't delete apps in the org, even those I've created myself, so I'm thinking this might not work?

lib/tasks/test.js
AlbertoElias: maybe remove the return new Promise and take a look into it when I look at how to solve exiting/erroring?

README.md
AlbertoElias: Done

lib/origami-build-tools.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 
 require('es6-promise').polyfill();
+require('isomorphic-fetch');
```
matthew-andrews: maybe `node-fetch` is a better fit here

matthew-andrews: maybe `node-fetch` is a better fit here

lib/tasks/test.js
matthew-andrews: most test frameworks understand promises should just be able to return the promise

lib/origami-build-tools.js
AlbertoElias: It's only here for haikro, why would node-fetch be better?

lib/tasks/test.js
AlbertoElias: I'm running the test via the command line, how can that interact with promises?

lib/tasks/install.js
```diff
@@ -9,8 +9,8 @@ var commandLine = require('../helpers/command-line');
 var files = require('../helpers/files');
 
 var versions = {
-	sass: '~3.3.0',
-	scssLint: '^0.27.0',
+	sass: '~3.4.0',
+	scssLint: '^0.28.0',
```
kaelig: Looks like this file is outdated

AlbertoElias: Yeah, don't worry about that. I will pull in master before merging

lib/tasks/test.js
AlbertoElias: poking @matthew-andrews

AlbertoElias:
I'm going to close this a re raise it to not make the history even worse

PR #234, opened 2015-03-26
Use this version of gulp-scss-lint 

samgiles:
It does not support synchronous mode, which caused an external dependency to be required, which required native compilation - sloooow stuff for stuff we don't even use.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "gulp-mustache": "^0.4.0",
     "gulp-rename": "^1.2.0",
     "gulp-ruby-sass": "^0.7.1",
-    "gulp-scss-lint": "^0.1.7",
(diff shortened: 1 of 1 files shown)


samgiles:
I'm gonna cheat and merge this as it's only a dependency update and significantly reduced install time :).

package.json
```diff
@@ -32,7 +32,7 @@
     "gulp-mustache": "^0.4.0",
     "gulp-rename": "^1.2.0",
     "gulp-ruby-sass": "^0.7.1",
-    "gulp-scss-lint": "^0.1.7",
+    "gulp-scss-lint": "samgiles/gulp-scss-lint",
```
AlbertoElias: can you add a semver range?

PR #235, opened 2015-03-27
Update browserify dependency

samgiles:
This has a dependency on a fork of `textrequireify`, waiting for that to get merged and published to NPM.  Don't merge before I've updated the `package.json` here for it.

This is a breaking change as it alters the API for configuring browserify transforms.

### DIFF
--- a/lib/tasks/build.js
+++ b/lib/tasks/build.js
@@ -14,6 +14,10 @@ var files = require('../helpers/files');
 var log = require('../helpers/log');
 var hash = require('gulp-hash');
 
+var textrequireify = require('textrequireify');
+var debowerify = require('debowerify');
+var babelify = require('babelify');
+
 var hashOptions = {
 	algorithm: 'sha1',
 	hashLength: 10,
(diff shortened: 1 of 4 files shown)


lib/tasks/build.js
```diff
@@ -33,19 +37,11 @@ module.exports.js = function(gulp, config) {
 	config = config || {};
 	var src = config.js || files.getMainJsPath() || null;
 	if (src) {
-		var transforms = ['babelify', 'debowerify', 'textrequireify'].concat(config.transforms || []);
-		delete config.transforms;
-
-		// Hackily shallow clone the config object
```
samgiles: @matthew-andrews  It seems you added this, I can't see an obvious reason to keep it.  Can you remember why it was needed exactly?

samgiles: @matthew-andrews  It seems you added this, I can't see an obvious reason to keep it.  Can you remember why it was needed exactly?

samgiles:
ftlabs/textrequireify#3 ftlabs/textrequireify#2

lib/tasks/build.js
matthew-andrews: i cannot, i'm sorry…

samgiles:
Ready for review!  

This is a breaking API change unfortunately, browserify transforms must be passed in as functions now, which I think I prefer anyway.

AlbertoElias:
But don't merge yet! We need to compare with webpack. If we're going to do a breaking change, we need to make sure it's the ideal one :)

matthew-andrews:
@AlbertoElias @samgiles re: breaking change — i think we could use this opportunity to remove the asset hashing from OBT.  Really sorry about this but asset hashing turns out to be really quite hard…  we have a mechanism that works now for Next but using home brewed tooling.

Asset hashing is actually two challenges — the easier of the two (calculating the hash, which is what OBT does — and can actually be achieved with a couple of lines of code) and storing a full history of all previous versions (which involves service architecture and state and urgh)

AlbertoElias:
@matthew-andrews Ok, we'll get rid of it! It's always nice having to remove things :)

PR #238, opened 2015-03-30
Adds troubleshooting guide and polishes semver in install task

AlbertoElias:


README.md
```diff
@@ -138,8 +137,6 @@ Local demos consist of HTML, CSS and JS (if Sass & JS exists), and are created i
 
 _(Sourcemaps aren't generated as this feature is incompatible with csso. We will revisit this when [gulp-ruby-sass](https://github.com/sindresorhus/gulp-ruby-sass) 1.0 is released)_
```
kaelig: Feel free to edit `csso` into `clean-css` (I'd do it but that might generate a merge conflict)

TROUBLESHOOT.md
```diff
@@ -0,0 +1,34 @@
...
+		- Note: rbenv needs a bit of post-install setup. If you install with homebrew, also do steps 2 and 3 of the [Basic GitHub Checkout](https://github.com/sstephenson/rbenv/#basic-github-checkout) section.
+
+### Error `Unable to download data from https://rubygems.org/`
+
+If you receive an error specifying `Unable to download data from https://rubygems.org/ - SSL_connect returned=1 errno=0 state=SSLv3 read server certificate B: certificate verify failed (https://api.rubygems.org/specs.4.8.gz)` you'll need to manually update your gem package using the directions in [this gist](https://gist.github.com/luislavena/f064211759ee0f806c88).
+
+### Error `While executing gem ... (Gem::FilePermissionError)`
+
+When installing Sass and/or ScssLint, depending on how you have your permissions configured in your computer, you might get this error. You will need to run the command manual prefixing sudo to it. To install the supported version you need to run the following command (substituting the version number by the one mentioned [here](https://github.com/Financial-Times/origami-build-tools/#install)):
```
kaelig: The copy of this paragraph is slightly too personal. When possible, delegate blame on the machine without mentioning the user themselves. E.g.

> …, depending on how permissions are set on the machine, `obt install` might fail and return "While executing gem ... (Gem::FilePermissionError)…".

Also, I believe we don't recommend sudo-ing anymore. Should we redirect to the "Installing Ruby" section instead?

AlbertoElias: Good point.

About recommending not sudo-ing, keeping in mind that it's usually installed by default, we can't really do much about it.

README.md
AlbertoElias: I feel extremely free to do it :P Good catch!

TROUBLESHOOT.md
```diff
@@ -0,0 +1,34 @@
...
+
+```bash
+sudo gem install scss-lint -v 0.34.0
+sudo gem install sass -v 3.4.0
+```
+
+### Warning `Origami registry is not configured in a .bowerrc file`
+
+The install task runs `bower install`, which needs to check the Origami registry apart from the bower one. That's why you need to configure a `.bowerrc` file. You can do this in your home directory by running the following command:
```
kaelig: > which needs to check the Origami registry apart from the bower one.

I find this ambiguous. Maybe clearer to say something like "Origami modules are listed on Origami's custom Bower registry, in opposition to the global, public one".

> That's why you need to configure a `.bowerrc` file. You can do this in your home directory by running the following command:

Too personal? What do you think of: "Custom registries can be added to a [`.bowerrc`](http://bower.io/docs/config/) configuration file located in your home directory. To create a `.bowerrc` configuration file now, run:…"

kaelig:
Looks good aside from the few comments

TROUBLESHOOT.md
AlbertoElias: Yup, much better, thanks!

TROUBLESHOOT.md
kaelig: Isn't that what this is supposed to fix?

> If you get a `Gem::FilePermissionError` error, you need to [install a ruby version manager](http://stackoverflow.com/questions/19579392/installing-gem-fails-with-permissions-error).

AlbertoElias: The error is:

> ERROR:  While executing gem ... (Gem::FilePermissionError)
    You don't have write permissions for the /Library/Ruby/Gems/2.0.0 directory.

kaelig: Isn't this is where users should either: `sudo gem install…` or install a ruby manager? If so, then should we favor the installation of a ruby version manager instead of sudo?

AlbertoElias: The error doesn't really say anything about a Ruby manager, but I'm not too familiar with Ruby. If you think there's a better solution, what would it be?

AlbertoElias:
@kaelig fixes pushed

kaelig:
Cheers, great stuff.

PR #239, opened 2015-03-30
JS -> CSS

kaelig:
Quick question:

> Also generates a JSON file with mappings to the original filename.

Does it?


AlbertoElias:
@kaelig Yup, it needs it for the mapping

PR #241, opened 2015-04-01
Remove asset hashing

samgiles:


AlbertoElias:
Remove the hash documentation for sass as well. Rest is fine

PR #246, opened 2015-04-16
update textrequireify

quarterto:
adds `require("text!file")` support back.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
     "portfinder": "^0.4.0",
     "sassdoc": "^2.1.2",
     "semver": "^4.3.3",
-    "textrequireify": "^2.0.1",
+    "textrequireify": "^2.1.0",
     "through2": "^0.6.2",
     "vinyl-buffer": "^1.0.0",
     "vinyl-source-stream": "^1.0.0",


matthew-andrews:
should be unnecessary

matthew-andrews:
the `^` should mean you should already get this update if you do `npm update textrequireify`

matthew-andrews:
:-1:

quarterto:
it's nice to have an explicit update, and there's precedent for it (#222)

matthew-andrews:
hmm only `gulp-lintspaces` was unnecessary in that example — both `gulp-webserver` and `babelify` were necessary in that case………

matthew-andrews:
obviously doesn't hurt but sort of a bad habit i think

quarterto:
Trouble with leaving `package.json` as-is is that to get the new `textrequireify` someone would have to reinstall OBT. Releasing a new version makes it explicit.

pornel:
I think it's necessary when Origami modules start depending on the newer textrequireify — then we'll want to be able to say "OBT vXX *definitely* works with this" (rather than "OBT may or may not work depending on when you've installed it")

quarterto:
:+1:

PR #254, opened 2015-04-28
Updates nightwatch to work with latest haikro

AlbertoElias:


lib/tasks/test.js
```diff
@@ -146,14 +147,14 @@ module.exports._runBrowserTest = function(config) {
 			configContent = JSON.parse(configContent);
 			configContent.test_settings.default.launch_url = config.testUrl;
 
-			fs.writeFile(nightwatchConfigPath, JSON.stringify(configContent), function(error) {
+			fs.writeFile(nightwatchConfigPath.replace('json', 'tmp.json'), JSON.stringify(configContent), 'utf8', function(error) {
```
samgiles: Not sure why this change is necessary, but could it be one once on line 138?

lib/tasks/test.js
```diff
@@ -181,7 +184,7 @@ module.exports.browserTest = function(gulp, config) {
 	return Promise.all([
 		process.env.HEROKU_AUTH_TOKEN ? Promise.resolve(process.env.HEROKU_AUTH_TOKEN) : commandLine.run('heroku', ['auth:token']),
 		commandLine.run('git', ['rev-parse HEAD']),
```
samgiles: Not part of the PR so don't worry too much, but it might be better to be consistent here. Instead of space separating arguments, having them as separate entries into the array.

lib/tasks/test.js
AlbertoElias: What change exactly? I don't want to modify the original file, so I write to a temporary one

lib/tasks/test.js
AlbertoElias: I'll change it here though

lib/tasks/test.js
samgiles: Makes sense, I just mean de-duplicate the logic on line 150 and 157 so you only do this in one place.

AlbertoElias:
@samgiles I've applied the changes

lib/tasks/test.js
```diff
@@ -146,14 +147,16 @@ module.exports._runBrowserTest = function(config) {
 			configContent = JSON.parse(configContent);
 			configContent.test_settings.default.launch_url = config.testUrl;
 
-			fs.writeFile(nightwatchConfigPath, JSON.stringify(configContent), function(error) {
+			var tempNightwatchConfigPath = nightwatchConfigPath.replace('json', 'tmp.json')
```
samgiles: There is a JSHint error here: missing semi colon.

samgiles:
Nice, :+1: merge when ready. :)

PR #256, opened 2015-04-29
Go back to using the official scss-lint gulp task

kaelig:
It now uses [sync-exec](https://www.npmjs.com/package/sync-exec), an fs.execSync replacement until you get it natively from node 0.12+.

@samgiles I think your fork is better than this version but I believe it's probably better to follow the main project instead of relying on our own forks.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "gulp-mustache": "^1.0.2",
     "gulp-rename": "^1.2.0",
     "gulp-ruby-sass": "^1.0.1",
-    "gulp-scss-lint": "samgiles/gulp-scss-lint#v1.0.0",
+    "gulp-scss-lint": "^0.1.12",
     "gulp-sourcemaps": "^1.5.1",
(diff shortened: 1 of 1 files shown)


AlbertoElias:
We're using node 0.12 now

kaelig:
Yep, sorry I didn't put that into context.

Sam's fork of gulp-scss-lint removes the dependency to execSync (see https://github.com/samgiles/gulp-scss-lint/commit/b716f65897416d238d8432176071af0bc89264d9) because it required native compilation. Now that gulp-scss-lint doesn't rely on it anymore (but on sync-exec instead) we can use the original module instead.

samgiles:
Nice spot, I agree following this project is ideal.  I still think it's got a lot of extra unnecessary stuff, but I know, deep down, I'm never going to maintain a version. :+1:

kaelig:
"deep down" :performing_arts: the drama!

PR #257, opened 2015-04-30
Libsass is in da house

AlbertoElias:
It's using a beta of node-sass and an alpha of gulp-sass. We can release a beta version of obt v3, and make sure everything's working fine while we wait for those two to go stable. Everything seems to be working fine though


samgiles:
:+1:  \o/

kaelig:
:+1:

PR #261, opened 2015-05-19
Build refactor

AlbertoElias:


README.md
```diff
@@ -87,7 +88,7 @@ Runs:
 * __js(gulp, config)__ Config accepts:
 	- js: `String` Path to your main JavaScript file. (Default: './main.js' and checks your bower.json to see if it's in its main key)
 	- buildJs: `String` Name of the built JavaScript bundle. (Default: 'main.js')
-	- buildFolder: `String` Path to directory where the built file will be created. (Default: './build/')
+	- buildFolder: `String` Path to directory where the built file will be created. If set to `null`, files won't be saved. (Default: './build/')
```
samgiles: Do you mean 'if undefined, files won't be saved'?  Otherwise it looks like you need to do the following in your config to skip saving the files.

```JS
{
  buildFolder: null
}
```

AlbertoElias: it's if null. I can't make it undefined because by default I want it to be saved in a default folder, and the default includes `undefined`, that's why I decided on null

samgiles: I had a similar issue in [`o-errors`](https://github.com/Financial-Times/o-errors/blob/master/test/test_errors.js#L81). I originally had the config key named `disabled` so setting it to a truthy value would disable the module, and any falsey value, including undefined, would leave the module enabled.

This meant, by default, if you omitted `disabled` from the config you'd get the default 'enabled' behaviour.

Andrew suggested the config be named 'enabled', which I kind of agreed with.  Except I wanted it to be enabled by default if it was undefined.  [The code for this logic is here](https://github.com/Financial-Times/o-errors/blob/master/src/js/oErrors.js#L77).

I'd really avoid using null to have any semantic meaning.

samgiles: I misread a little there.  You want three values here.  Default, specified location, or disabled.  I'd introduce a new config key for the final, disabled, option.

AlbertoElias: I don't mind calling it a string or passing it null. I chose null preventing the extreme edge case of having a directory called disabled. Why is it not a good idea to give null some context?

samgiles: Hmm, for me it's that `null` ends up having meaning that it wasn't intended to have, and that the meaning is only valid in this one particular context if you've scrutinised the documentation.  It's one of those gut feelings that this is wrong thing to do.

I'll leave it up to you though.

AlbertoElias: changed

samgiles:
LGTM :+1:

Issue #259, opened 2015-05-06
More intelligent watch


matthew-andrews:
Would it be possible to make the watch a little more intelligent?

e.g. changes to javascript only rebuilds javascript, scss changes only trigger css changes

Feels like this would be quite a simple change.

matthew-andrews:
cc @wheresrhys

matthew-andrews:
this is how we achieved this in nbt https://github.com/Financial-Times/next-build-tools/blob/master/tasks/build.js#L31-36

AlbertoElias:
Fixed in #263 

Can we close?

matthew-andrews:
looks good to me!

PR #263, opened 2015-05-20
Made watch more intelligent by only rebuilding js or sass in build an…

AlbertoElias:
…d demo when one of those type of files has changed

I'm not sure the extra code hassle is worth it, or if this is the best way to do it. What do you think @samgiles ? It's definitely faster

### DIFF
--- a/lib/tasks/verify.js
+++ b/lib/tasks/verify.js
@@ -177,5 +177,5 @@ module.exports.origamiJson = function() {
 	return result;
 };
 
-module.exports.watchable = true;
+module.exports.watchable = false;
 module.exports.description = 'Lint code and verify if module structure follows the Origami specification';
--- a/lib/tasks/build.js
+++ b/lib/tasks/build.js
@@ -21,8 +21,12 @@ var debowerify = require('debowerify');
 var babelify = require('babelify');
 
 module.exports = function(gulp, config) {
-	var jsStream = module.exports.js(gulp, config);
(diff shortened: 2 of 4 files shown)


AlbertoElias:
Closes #259

lib/tasks/verify.js
```diff
@@ -177,5 +177,5 @@ module.exports.origamiJson = function() {
 	return result;
 };
 
-module.exports.watchable = true;
+module.exports.watchable = false;
```
samgiles: Why change this to false?

AlbertoElias: It's not really useful as the watching stops as soon as there's a linting
error. I implemented the cleverer watch there, and then figured out that
verify --watch is just not useful

-- 

number 53723.

samgiles:
Yeh the extra code is annoying but with this architecture is the simplest thing to do. I think there's probably a cleaner way to do it though: it feels like 'watch' should be a trait of a task rather than a task itself.

Maybe you could invoke, for example, the `js` build task with a `watch` option and it task itself will register a `watch` on only the files that task is interested in for example.

This is something to consider I think, but for now this PR LGTM, especially as it makes watch quicker.

AlbertoElias:
I thought about that, but I think that might end up requiring a lot of repeated boilerplate code

samgiles:
imo boilerplate code ==  good code :D.   I honestly think it would be fine.

PR #266, opened 2015-05-28
Fix typo within npm version notice.

phamann:
After recently going through the install process and having this specific issue, I can vouch that the problem occurs with npm > 2.7.5.


AlbertoElias:
Oh yes! Thanks!

PR #270, opened 2015-06-15
CLI won't break if task does not return promise or stream, and docume…

AlbertoElias:
…nts runServer instead of local


kaelig:
:+1:

PR #274, opened 2015-06-18
Suppress stderr when getting the version of a command

matthew-andrews:
I can't think of a case where wanting the stderr of a `--version` call would ever be useful……

Also because ruby is corrupted on Next's Jenkins machine (cc @captain-sysadmin) this means OBT doesn't work for any of our Stash repos :-(.


matthew-andrews:
So in this case although I know it's our fault for not installing ruby properly but I think this would make OBT far more resilient slightly broken setups.

PR #276, opened 2015-06-23
Show message after stream tasks and cleans up code

AlbertoElias:


lib/origami-build-tools-cli.js
```diff
@@ -96,6 +96,8 @@ if (tasks.isValid(argument)) {
 				taskResult
 					.on('error', reportTaskError)
 					.on('end', reportFinished);
+				// Make sure it reaches 'end'
+				taskResult.resume();
```
samgiles: To be defensive, you might want to add to the condition on line 95: `if (taskResult.on && taskResult.resume)`, we should probably also handle the case where we don't understand the returned value from the task, but that can be in another PR.

AlbertoElias: The only one we don't understand is version, and we don't need to do anything, and if we don't understand the returned value we shouldn't do anything, right? I also think all tasks should return a promise or a stream and I would rather make version return a promise (though it's probably an overkill)

samgiles: Ok, sure, but in this instance, takResult could define `on` but not `resume` and cause a 'fatal' error, depends on the implementation of the stream being returned.

samgiles:
Apart from the test failure it looks good, are the test failures related?

lib/origami-build-tools-cli.js
AlbertoElias: I have already pushed a change to that a while ago. I'm trying to find out why the build gets stuck

AlbertoElias:
@samgiles I think it's now fixed

gulpfile.js
```diff
@@ -11,7 +11,11 @@ gulp.task('verify', function() {
 });
 
 gulp.task('test', function() {
-	obt.test.npmTest(gulp);
+	obt.test.npmTest(gulp)
```
samgiles: You can return a Promise or stream within `gulp.task`, would this set up error handling for you? https://github.com/gulpjs/gulp/blob/master/docs/API.md#return-a-promise

lib/tasks/demo.js
```diff
@@ -199,8 +193,9 @@ function hasUniqueNames(demos) {
 
 function updateOrigamiJsonDemoList(demosBuilt) {
 	var origamiDemoConfig = demosBuilt.map(function(conf) {
+
 		return {
-			path: "/" + conf.outputHtmlPath,
+			path: '/demos/' + conf.name + '.html',
```
samgiles: Should really use `path.join` to build paths, it's more portable (handling special cases for windows), and it's safer.

AlbertoElias: Yup!

gulpfile.js
AlbertoElias: Do you mean to return the function instead of just running it?

lib/tasks/demo.js
AlbertoElias: Though it's not really that important because I just want to generate the string, that path isn't going to be used anywhere

AlbertoElias:
@samgiles done :)

gulpfile.js
samgiles: No I mean return the result of running the function.

PR #282, opened 2015-06-26
Desktop notifications and better reporting from `--watch`

samgiles:
When using watch in the background, you'll get cross platform desktop notifications!  Also improves the success message on the cli output.  

screen shot 2015-06-26 at 15 46 41
screen shot 2015-06-26 at 15 47 50

Thoughts?


lib/tasks/watch.js
```diff
@@ -20,6 +23,28 @@ var watchGlobs = [
 	'!**/tmp-src.scss'
 ];
 
+function reportTaskComplete(taskResult, builtExtensions) {
+
+	checkTaskStatus(taskResult, function(err, result) {
+
+		var notificationTitle = "Build succeeded";
```
AlbertoElias: The messages are only valid for the build task, not for demo or test which are the other watchable tasks

lib/tasks/watch.js
```diff
@@ -30,7 +55,7 @@ module.exports = {
 
 		watcher.on('ready', function() {
 			log.secondary('Running tasks...');
-			task(gulp, config);
+			reportTaskComplete(task(gulp, config), ['js', 'css']);
```
AlbertoElias: passing that extra info is only valid for the build and demo task

lib/tasks/watch.js
```diff
@@ -41,7 +66,7 @@ module.exports = {
 			} else if (fileExtension === '.scss') {
 				config.watching = 'sass';
 			}
-			task(gulp, config);
+			reportTaskComplete(task(gulp, config), [ config.watching ]);
```
AlbertoElias: config.watching could be undefined

AlbertoElias:
I'm going to close this as it's probably a bit too overwhelming to have notifications for these things

PR #283, opened 2015-06-26
Added custom reporter as the integrated one stopped working and proje…

AlbertoElias:
…ct is not actively maintained. Also made output cleaner


samgiles:
lgtm :+1:!

PR #284, opened 2015-06-26
Upgrade all deps

AlbertoElias:


samgiles:
:+1: looks good, should our official releases contain a shrinkwrap?  I know we've not come across any issues with that yet so might not be an issue.

AlbertoElias:
I prefer semver ranges as long as they're safe

PR #286, opened 2015-07-02
Allow the cwd to be configurable for the build.js task.

samgiles:


AlbertoElias:
Why not merge to a 'build-service' branch so you can push things quickly and that way we don't need to make releases regularly?

samgiles:
Ok, I'll just rename this branch.

PR #290, opened 2015-07-07
Parameterisation changes required for integration into the build service.

samgiles:


samgiles:
oops I need to fix those tests..

PR #292, opened 2015-07-14
Replace JSHint with eslint

imranolas:
There has already been some conversation about moving to eslint on https://github.com/Financial-Times/ft-origami/issues/375 and hopefully this goes some way to implementing it as per that discussion.  

From our perspective on NEXT, the limited configurability of JSHint isn't just inconvenient but restrictive as it is a required dependency for next-build-tools.


AlbertoElias:
I have been looking into implementing and have it running locally. I am not pushing it up until we decide on the config that I posted in that issue. Once that's done, I'll push it up. JSHint isn't being removed, as that would be a breaking change. If we can agree on the config, I would like to make the pr tomorrow which is basically what you've done as well

AlbertoElias:
I opened #293 We should probably move the conversation over there. We finally decided to add a deprecation warning to JSHint in 2.x and remove JSHint as well.

PR #293, opened 2015-07-15
Removes jsHint in favour of ESLint. Adds Babel-Runtime polyfill and t…

AlbertoElias:
…ransformer to fix paths

DO NOT MERGE

So this would be a new major release. We would get Babel polyfills, and change to ESLint, which changes linting rules which will affect, probably, all projects. Rules should be discussed before merging. Tests are passing, but not verify as I would like to finalise the linting rules for both obt, and modules.

@triblondon @matthew-andrews


triblondon:
The syntax standard in the Origami spec will also need to be updated.

AlbertoElias:
Pinging @triblondon @matthew-andrews to check new linting rules

lib/plugins/babelRuntimePathResolver.js
```diff
@@ -0,0 +1,50 @@
...
+var falafel = require('falafel');
+var fs = require('fs');
+
+var prefix = /babel-runtime/;
+
+function create(config) {
+	return function(file) {
+		if (!(/\.js$/).test(file)) return through();
+
+		var rootDirectory = path.resolve(path.normalize(config.rootDirectory || './node_modules'));
```
samgiles: What is `./node_modules` relative to here?

samgiles:
See comment, otherwise :+1:

lib/plugins/babelRuntimePathResolver.js
AlbertoElias: It's relative to that file. I agree throwing would be better

IanVS:
Please note, [eslint 1.0.0](http://eslint.org/blog/2015/07/eslint-1.0.0-released/) and gulp-eslint 1.0.0 have been released in the last few days.  There are some breaking changes, but also many bug fixes.  Let me know if you have any questions about ESLint and I'll do my best to answer.

AlbertoElias:
@IanVS Thanks a lot for letting us know! I've just pushed the version

@samgiles Fixed that, and also fixed linting issues. I'm going to make a v4 beta release

test/fixtures/o-test/.eslintrc
```diff
@@ -0,0 +1,38 @@
...
+    "no-extend-native": 2,
+    "wrap-iife": 2,
+    "new-cap": 2,
+    "no-caller": 2,
+    "no-irregular-whitespace": 1,
+    "no-trailing-spaces": 2,
+    "no-multi-str": 0,
+    "dot-notation": 0,
+    "strict": [2, "function"],
+    "quotes": [1, "single"],
```
matthew-andrews: oh man……

test/fixtures/o-test/.eslintrc
```diff
@@ -0,0 +1,38 @@
...
+    "valid-jsdoc": 1,
+    "no-multi-spaces": 2,
+    "no-multiple-empty-lines": 2,
```
matthew-andrews: this feels like the job of lintspaces not a javascript linter…

test/fixtures/o-test/.eslintrc
```diff
@@ -0,0 +1,38 @@
...
+    "browser": true
+  },
+  "rules": {
+    "eqeqeq": 2,
+    "guard-for-in": 2,
+    "no-extend-native": 2,
+    "wrap-iife": 2,
+    "new-cap": 2,
+    "no-caller": 2,
+    "no-irregular-whitespace": 1,
```
matthew-andrews: leave to lintspaces?

test/fixtures/o-test/.eslintrc
```diff
@@ -0,0 +1,38 @@
...
+    "no-trailing-spaces": 2,
```
matthew-andrews: again, leave to lintspaces?

test/fixtures/o-test/.eslintrc
IanVS: Interesting, I hadn't seen lintspaces before.  Does it provide features/rules which ESLint does not?

matthew-andrews: lintspaces is just about whitespace… and is for more than just javascript…… works with any type of file

IanVS: Very neat, thanks.

AlbertoElias:
@matthew-andrews I agree with leaving things to lintspaces, I just tried to copy rules over and then open up the discussion

test/fixtures/o-test/.eslintrc
AlbertoElias: Very much in favour of removing the warning, it's quite noisy. I wanted to test it out, it's quite nice to have some kind of convention on modules, but after running verify on a few modules, it spilled out dozens and dozens of warnings. I'll probably keep it in obt though

IanVS: Not sure if it's helpful, but if you want to keep the warnings but not see them all the time, you can run ESLint with `--quiet` which will prevent display of warnings.  

Also, if you're getting a lot of errors/warnings, I made [eslint-nibble](https://github.com/IanVS/eslint-nibble) to help sort through them and make it less overwhelming.

matthew-andrews:
Oops, thanks :)

test/fixtures/o-test/.eslintrc
AlbertoElias: @IanVS regarding `--quiet`, what's the difference between that and just removing the rule? And also, does that option exist in `gulp-eslint`, doesn't seem to

IanVS: Ah that's right, forgot you are using gulp-eslint.  I haven't used that, I typically just create an npm script and run eslint before my tests that way. 

The difference with quiet would be that you could still run without it and see the output without having to change the config, which is useful if it's your goal to clean it up eventually.

lib/tasks/verify.js
```diff
@@ -149,23 +149,23 @@ module.exports.scssLint = function(gulp, config) {
...
 	var combinedStream = combine(
 		gulp.src(['**/*.js'].concat(excludeFiles)),
-		jshint(configPath),
-		jshint.reporter('default'),
-		jshint.reporter('fail')
+		eslint({
+			configFile: configPath
+		}),
+		eslint.format('compact', process.stdout),
+		eslint.failAfterError()
```
matthew-andrews: hmm we sure we want this?

>         // To have the process exit with an error code (1) on
>        // lint error, return the stream and pipe to failOnError last.

if you're pulling this in as a dependency and running it from inside node this behaves very oddly…  is this how it worked in jshitn?

AlbertoElias: What do you mean it behaves oddly? What's the behaviour you would expect?

test/fixtures/o-test/.eslintrc
AlbertoElias: Right, thanks, I think I'm going to get in less fights if I just remove the rule altogether.

test/fixtures/o-test/.eslintrc
AlbertoElias: Doesn't have a rule for this

test/fixtures/o-test/.eslintrc
AlbertoElias: yup

test/fixtures/o-test/.eslintrc
AlbertoElias: I'm not so sure about moving this to lintspaces. Lintspaces errors with /* */ comments saying they're new lines even though I have the ignore js-comments rule

AlbertoElias: Actually, it's only an issue with multiline comments

```js
/*
*
*/
```

This `/* comment */` is fine

AlbertoElias:
I've released a new beta, rules should be final, only concern is [this](https://github.com/Financial-Times/origami-build-tools/pull/293#discussion_r36639881)

test/fixtures/o-test/.eslintrc
samgiles: That's a bug with lintspaces,  the latest version of lintspaces definitely has a regex that matches that string you've got there.

https://github.com/schorfES/node-lintspaces/blob/master/lib/constants/ignorePatterns.js#L3

config/.eslintrc
```diff
@@ -0,0 +1,37 @@
...
+		"prefer-const": 1,
+		"no-const-assign": 2
+	},
+	"globals": {
+		"require": false,
+		"module": false,
+		"exports": false,
+		"requireText": false
+	}
+}
```
triblondon: Is this the same as config/.eslintrc?  If so do we have to duplicate?

AlbertoElias: No, it has a difference, that way we can test that it's using the custom config instead of the default one in its corresponding test

PR #300, opened 2015-08-13
Improves install errors messages and pushes babel-core version which …

AlbertoElias:
…had a bug with const in for of loop

### DIFF
--- a/TROUBLESHOOT.md
+++ b/TROUBLESHOOT.md
@@ -77,7 +77,9 @@ When installing [bower](https://bower.io), depending on how permissions are set 
 >npm ERR!
 >npm ERR! Please try running this command again as root/Administrator.
 
-The command can be ran manually prefixing sudo to it. It's the following command (substituting the version number by the one mentioned [here](https://github.com/Financial-Times/origami-build-tools/#install)):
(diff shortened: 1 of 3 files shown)


TROUBLESHOOT.md
```diff
@@ -77,7 +77,9 @@ When installing [bower](https://bower.io), depending on how permissions are set
 >npm ERR!
 >npm ERR! Please try running this command again as root/Administrator.
 
-The command can be ran manually prefixing sudo to it. It's the following command (substituting the version number by the one mentioned [here](https://github.com/Financial-Times/origami-build-tools/#install)):
+This can be fixed by following one of the methods suggested in the [official guide](https://docs.npmjs.com/getting-started/fixing-npm-permissions).
+
+If the error still occurs, the command can be ran manually prefixing sudo to it. It's the following command (substituting the version number by the one mentioned [here](https://github.com/Financial-Times/origami-build-tools/#install)):
```
samgiles: Should we just remove this section, it's rather unsafe to suggest it.

AlbertoElias: Well, it is how these things get installed by default, and it's an easy workaround. For non-devs, it's probably not that much of a security risk and it probably helps give a more seamless experience

PR #302, opened 2015-08-31
Webpack

AlbertoElias:
So there are currently two issues.

1. Textrequireify has been made into a custom loader, but I need to get the path of the file that's being modified and that's not working. I asked in [stack overflow](http://stackoverflow.com/questions/32320038/cant-access-resourcepath-property-from-webpack-loader) (following the webpack contributing guidelines) to see if I'm doing something wrong. The other alternative is to use a transform loader like the web app team and just use the original code, but we need to pass in the cwd as a config option, and the transform loader doesn't support passing in custom parameters. Check out [this issue](https://github.com/webpack/transform-loader/pull/8)

2. Webpack doesn't act the same way as browserify and still builds when it comes upon certain errors like unmet dependencies. Need to figure out a way around this. @samgiles check out [this](https://github.com/Financial-Times/origami-build-tools/compare/eslint...webpack?expand=1#diff-5a41c7b9bce0c7c00e0f564f21cb7603R92)


antoligy:
Unfortunately this beta currently breaks our build process as bower-webpack-plugin does not respect the directory setting in .bowerrc.

I've created lpiepiora/bower-webpack-plugin#25 addressing this, and will attempt to contribute a PR addressing that. (If that gets declined, or isn't merged in time, happy to contribute similar here.)

antoligy:
Submitted piepiora/bower-webpack-plugin#27.

antoligy:
Submitted lpiepiora/bower-webpack-plugin#27.

antoligy:
Unfortunately with that applied there are still some failures, due to the hardcoded `bower_components` in the OBT build task.

antoligy:
Should be resolved in #310. :smile:

AlbertoElias:
This is ready to review. @triblondon could you take a look?

lib/origami-build-tools-cli.js
```diff
@@ -65,7 +65,7 @@ var argument = argv._[0];
 
 function reportTaskError(error) {
 	if (Array.isArray(error)) {
-		log.primaryError(error.join('\n'));
+			log.primaryError(error.join('\n'));
```
triblondon: weird indent?

lib/origami-build-tools.js
```diff
@@ -1,6 +1,5 @@
 'use strict';
 
-require('es6-promise').polyfill();
```
triblondon: So we now depend on Node 0.12 minimum?  Are we expressing that anywhere?

AlbertoElias: Node v4. In the package.json. But it works fine with Node 0.12 as well.

triblondon:
Did a review with Alberto.  I suggest capturing more of the gotchas into file and line comments, otherwise looks good.

PR #303, opened 2015-09-02
adding ability to specify browserify externals

wheresrhys:
Allows creation of a core bundle to share across next apps

### DIFF
--- a/lib/tasks/build.js
+++ b/lib/tasks/build.js
@@ -56,9 +56,21 @@ module.exports.js = function(gulp, config) {
 
 		log.secondary("Browserifying " + src);
 
+		var externals;
+		if (config.externals) {
+			config.bundleExternals = false;
+			externals = config.externals;
+			config.fullPaths = true;
+			delete config.externals;
+		}
+
 		var bundle = browserify(config)
 			.add(src)
 			.require(src, { entry: true });
+
+		if (externals) {
+			bundle = bundle.external(externals);
+		}
 
 		bundle = transforms.reduce(function(bundle, transform) {
 			if (transform.options && transform.transform) {


lib/tasks/build.js
```diff
@@ -56,10 +56,22 @@ module.exports.js = function(gulp, config) {
 
 		log.secondary("Browserifying " + src);
 
+		var externals;
+		if (config.externals) {
+			config.bundleExternals = false;
+			externals = config.externals;
+			config.fullPaths = true;
+			delete config.externals;
```
AlbertoElias: why do you delete it?

AlbertoElias:
What do you mean by core bundle? What exactly are you passing in?

You are missing the readme bit, but keep in mind we're moving to webpack very shortly, so this won't work in v4

lib/tasks/build.js
wheresrhys: because it's not a valid config option for browserify and I don't want to risk leaving it in

wheresrhys:
This https://github.com/Financial-Times/n-core-js-bundle/blob/master/bundler.js. All bower components that are dependencies of all/most next apps

wheresrhys:
regarding webpack, would you mind adding this feature to the requirements for moving to webpack. We will definitely want to use it (it will mean our apps share one, heavily cached 60kb-ish js file, and each page will then typically only need to download a very small bundle of application code)

### Statement of the requirement:

Should support importing/requiring modules (including subdependencies) from a separate js bundle
e.g. if core.js contains

```
require('dep1')
````

and dep1.js contains

```
require('dep2')
````
and main.js contains

```
require('dep1')
require('dep2')
````
then it should be possible to build bundle.core.js and bundle.main.js in such a way that bundle.core.js contains application code for dep1 and dep2, but bundle.main.js requires dep1 and dep2 from bundle.core.js and doesn't contain any application code for dep1 or dep2

wheresrhys:
I think I'll focus on the tasks I need to complete to be able to deliver a separate core bundle and leave actually building it up to obt and webpack as doing it with browserify is quite complex and probably not worth the effort if we're changing implementations soon anyway

wheresrhys:
And finally, I notice our implementation of source maps for minified code is in next build tools (I thought it had made it's way into obt). It'd be good if you could test if this still works with webpack

wheresrhys:
I'll close this and open a couple of issues

PR #306, opened 2015-09-10
Better sudo explanation

### LINKED ISSUE #305: Find safer instructions for installing OBT globally
image

Look, even I know it's bad to tell origami users (who may not be power users) to blindly use sudo. Sure it's okay for OBT, but the implication is it's safe to use sudo to install any old node package. And it really, really isn't. Right?

http://bfy.tw/1hzs

AlbertoElias:
Fixes #305

### DIFF
--- a/lib/helpers/update-notifier.js
+++ b/lib/helpers/update-notifier.js
@@ -53,8 +53,8 @@ function updateNotifier() {
 			var currentVersion = getCurrentVersion();
 			if (semver.lt(currentVersion, latestVersion)) {
 				console.log(String('Origami Build Tools update available: ').bold + String(latestVersion).bold.green + ' ' + String('(current: ' + currentVersion + ')').grey);
-				console.log(String('Run ').bold + String('npm install -g origami-build-tools').bold.blue + String(' to update.').bold);
-				console.log(String('You may need to run this command with sudo'));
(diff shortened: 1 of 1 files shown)


adambraimbridge:
Thanks for looking into this. Is there a nice URL you can add to the message to point to the npm setup instructions? 

image

AlbertoElias:
@adambraimbridge how about now?

adambraimbridge:
Unfortunately that guide just has more instructions using sudo, so now it's like stepping out of the frying pan into the fire :) 

I guess the problem's bigger than I thought. _Looks like it's impossible for a n00b to get up and running to develop new sites based on origami without using sudo._ So the implication that it's safe to use sudo is still there. 

Maybe the only way to make it better is to keep stressing to n00bs how important it is to be careful when using sudo ...

Anyway, :+1:  for this PR :)

AlbertoElias:
Actually, we only recommend sudo after offering a way to configure your permissions so you don't need sudo for both npm and Ruby Gems. There is one place where we still only suggest sudo, and I'll update that right away

AlbertoElias:
I've updated the troubleshoot guide removing that use of sudo, and adding a notice about the risk of using sudo

adambraimbridge:
Excellent. Thanks again Alberto :+1:

AlbertoElias:
Thank you for reporting!

PR #314, opened 2015-09-25
Starts move to es6 by passing new linting rules

AlbertoElias:
I've checked it passes test and verify, and I've also tried to run all the tasks to make sure everything works, hopefully I didn't miss anything


expe:
Are the 'use strict' statements still necessary in ES6 modules? We have `"strict": [2, "global"]`, so this errors for me.

Looks fine otherwise.

AlbertoElias:
This is Node and there are no ES6 modules, it's still CommonJS modules, so use strict is necessary. The eslint rules for obt are different thant the ones for Origami modules, and you can run the OBT ones by running `gulp verify`

expe:
Looks good :+1:

PR #320, opened 2015-10-07
obt demo now uses origami.json config

AlbertoElias:


test/tasks/demo.test.js
```diff
@@ -95,11 +95,22 @@ describe('Demo task', function() {
...
-					fs.unlink('demos/src/config.js');
+				});
+		});
+
+		it('should not fail using origami.json', function(done) {
+			demo(gulp, {
+				demoConfig: 'origami.json'
+			}).catch(function(err) {
+				setTimeout(function() {
+					expect(err).to.not.be('Couldn\'t find demos config path, checked: origami.json');
```
triblondon: So this test passes if the demo task throws an error that doesn't match the string you're checking?

triblondon:
Slightly confused by test wording - one comment there otherwise lgtm.

test/tasks/demo.test.js
AlbertoElias: Yeah, tests for demos aren't super great, but it's like that on all of them, they're missing a template for another test, so they throw that error. I'll add in a comment

PR #321, opened 2015-10-09
Adds support for only building one demo so the build service can use it

AlbertoElias:


lib/tasks/demo.js
```diff
@@ -258,8 +258,10 @@ module.exports = function(gulp, config) {
 		}
 
 		demosConfig.demos.forEach(function(demoConfig) {
-			// Extend an empty object to avoid changing the value of defaultDemoConfig
-			demos.push(extend(true, {}, defaultDemoConfig, demosConfig.demosDefaults || demosConfig.options, demoConfig));
+			if (!config.buildDemo || (config.buildDemo && config.buildDemo === demoConfig.name)) {
```
triblondon: Not sure on the naming (sorry). `buildDemo` feels like an action or a task to me, not a config option.

All OBT options are global, right, not scoped to their task?  Could we call it `demoFilter`, maybe?

Also needs adding to README.

AlbertoElias: I don't like the name either, I thought it would be better to just take the conversation public. Not sure about demoFilter either, might be good if it accepts an array instead of a string. I'll implement that

AlbertoElias: I didn't add it to the readme because I thought it was only a build service thing and better to be kept private

AlbertoElias:
Updated @triblondon

triblondon:
:+1:

PR #323, opened 2015-10-14
Building source maps using webpack

wheresrhys:
This builds a source map as a separate file, with development file paths hard coded so far.

I'm gonna work on creating what we need for next. Please leave comments about the requirements for the wider origami community in here and I'll try to accommodate both @AlbertoElias

### DIFF
--- a/lib/tasks/build.js
+++ b/lib/tasks/build.js
@@ -102,8 +102,11 @@ module.exports.js = function(gulp, config) {
 					}
 				]
 			},
+			devtool: useSourceMaps ? 'source-map' : undefined,
 			output: {
-				filename: dest
+				devtoolModuleFilenameTemplate: '[absolute-resource-path]',
+				filename: dest,
+				sourceMapFilename: dest + '.map'
 			}
 		};
 
@@ -114,14 +117,8 @@ module.exports.js = function(gulp, config) {
 		const combinedStream = combine.obj(
 			gulp.src(src),
 			webpack(webpackConfig),
-			gulpif(useSourceMaps, function() {
-				return sourcemaps.init({loadMaps: true})
-			}),
 			gulpif(config.env === 'production', function() {
 				return streamify(uglify());
-			}),
-			gulpif(useSourceMaps, function() {
-				return sourcemaps.write();
 			}),
 			gulpif(destFolder !== 'disabled', function() {
 				return gulp.dest(destFolder);
--- a/lib/tasks/build.js
+++ b/lib/tasks/build.js
@@ -3,13 +3,12 @@
 const path = require('path');
 const merge = require('merge-stream');
 const BowerPlugin = require('bower-webpack-plugin');
-const webpack = require('webpack-stream');
+const webpack = require('webpack');
+const webpackStream = require('webpack-stream');
 const combine = require('stream-combiner2');
-const streamify = require('gulp-streamify');
 const sass = require('gulp-sass');
 const sourcemaps = require('gulp-sourcemaps');
 const cleanCss = require('gulp-minify-css');
-const uglify = require('gulp-uglify');
 const rename = require('gulp-rename');
 const gulpif = require('gulp-if-else');
 const autoprefixer = require('gulp-autoprefixer');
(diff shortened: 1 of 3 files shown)


lib/tasks/build.js
```diff
@@ -69,6 +68,13 @@ module.exports.js = function(gulp, config) {
 
 		log.secondary('Webpacking ' + src);
 
+		const plugins = [bowerPlugin];
+
+		if (config.env !== 'development') {
```
wheresrhys: maybe this should be `=== 'production'` ?

AlbertoElias: yup

lib/tasks/build.js
```diff
@@ -3,13 +3,12 @@
 const path = require('path');
 const merge = require('merge-stream');
 const BowerPlugin = require('bower-webpack-plugin');
-const webpack = require('webpack-stream');
+const webpack = require('webpack');
+const webpackStream = require('webpack-stream');
```
AlbertoElias: you're now requiring both webpack and webpack-stream? =/ This slows down the process, requires are the main bottleneck for obt

lib/tasks/build.js
```diff
@@ -3,13 +3,12 @@
...
 const combine = require('stream-combiner2');
-const streamify = require('gulp-streamify');
 const sass = require('gulp-sass');
 const sourcemaps = require('gulp-sourcemaps');
 const cleanCss = require('gulp-minify-css');
-const uglify = require('gulp-uglify');
```
AlbertoElias: You can probably remove this from the package.json, not sure about streamify, double check that one

AlbertoElias:
Sass sourcemaps would still be inlined and that would be inconsistent. Why is it important for source maps to be in separate files? It seems more comfortable if only two files are output

lib/tasks/build.js
wheresrhys: yep, checked both

wheresrhys:
Performance. If inlined then every user in prod downloads them. If not then only users with dev tools open do.

AlbertoElias:
I thought sourcemaps was only for development

lib/tasks/build.js
wheresrhys: webpack was already in package.json, so presumably being required somewhere else already. But either way, I just checked the webpack-stream api and it does expose the webpack library as a property so will change to use this instead

AlbertoElias: I don't know where webpack is being used, maybe it can be removed from the package.json and that's a mistake. Having it in the package.json slows down install, but having it here slows down the actual build

wheresrhys: actually no it wasn't. so the `require('webpack')` should've errored. Are prod builds covered by tests?

wheresrhys:
Somehow (I don't understand the mechanism) raven is able to use the source map to send better information to sentry. It's also useful for debugging bugs that are difficult to reproduce except in prod

lib/tasks/build.js
AlbertoElias: Well, if you have it installed locally it won't error, and I'm not aware of
how to do a prod build test. In ci it should have failed

-- 

number 53723.

wheresrhys:
Also arguably sass sourcemaps should e separate files too

wheresrhys:
> can you just call it webpack like before?

That would be confusing I feel. e.g. when trying to use the uglify plugin the 'webpack' var name implied 'this is webpack' so I expected that I could use examples from the webpack docs exactly as they were. What's the benefit of storing webpack-stream in a variable called webpack?

AlbertoElias:
Yeah, I think it should be the same for Sass and Webpack.

Regarding the name, they're basically the same thing, I find webpack to be less confusing because what you're running is webpack, and I don't want to care what kind of object it is.

ironsidevsquincy:
plus one for all sourcemaps in separate files

wheresrhys:
So my proposal is to make sourcemaps in separate files the default, but to allow users to configure to be same file if they want to. I'm not intending to write the code for sass to do this as I don't think it's of any benefit to next, but am happy to do this for js and add a warning for sass to say that sourcemaps are inlined for the time being.

Sound ok?

AlbertoElias:
Just a small minor readme issue

PR #324, opened 2015-10-16
Make OBT output errors when --watch is used if an error exists

### LINKED ISSUE #318: Demo with --watch doesn't error or exit when there's an error
When there's a Sass error:

Current it hangs on `compiling`.
Expect it to error and exit.

It works as expected without the --watch param.

expe:
Removed 'watch module' and moved code to 'origami-build-tools-cli' so it can make use of the task checker.

Fixes #318 

CC @triblondon


AlbertoElias:
I would make it slightly more explicit by running the task before passing it to checktaskresult, so, like before, it's assigned to variable, and then taskResult is passed into the new function

Issue #325, opened 2015-10-18
obt verify currently broken


matthew-andrews:
A fresh install of all the latest dependencies produces this…  Started noticing this on Friday…
```
obt verify
events.js:141
      throw er; // Unhandled 'error' event
      ^

TypeError: Invalid non-string/buffer chunk
    at chunkInvalid (/Users/mandrews/sandboxes/next-ads-component/node_modules/origami-build-tools/node_modules/stream-combiner2/node_modules/readable-stream/lib/_stream_readable.js:421:10)
    at readableAddChunk (/Users/mandrews/sandboxes/next-ads-component/node_modules/origami-build-tools/node_modules/stream-combiner2/node_modules/readable-stream/lib/_stream_readable.js:176:12)
    at DuplexWrapper.Readable.push (/Users/mandrews/sandboxes/next-ads-component/node_modules/origami-build-tools/node_modules/stream-combiner2/node_modules/readable-stream/lib/_stream_readable.js:162:10)
    at Transform.<anonymous> (/Users/mandrews/sandboxes/next-ads-component/node_modules/origami-build-tools/node_modules/stream-combiner2/node_modules/duplexer2/index.js:50:15)
    at emitOne (events.js:77:13)
    at Transform.emit (events.js:169:7)
    at readableAddChunk (_stream_readable.js:146:16)
    at Transform.Readable.push (_stream_readable.js:110:10)
    at Transform.push (_stream_transform.js:128:32)
    at afterTransform (_stream_transform.js:76:12)
make: *** [test] Error 1
```

matthew-andrews:
Potentially casued by new version of stream-combiner
https://github.com/substack/stream-combiner2/releases

expe:
Noticing this on travis too.
https://travis-ci.org/Financial-Times/o-typography/builds/86151934

AlbertoElias:
I'll pay attention to stream-combiner2 updates

matthew-andrews:
So good to have you back Alberto :smile:

AlbertoElias:
Everything broke!

PR #326, opened 2015-10-18
Fix stream-combiner at 1.0.2

matthew-andrews:
‘Fixes’ #325


expe:
This is probably a temporary fix but it works for me :+1:

PR #328, opened 2015-10-22
document eslint fix

wheresrhys:


wheresrhys:
actually --fix seems pretty impotent when it comes up against almost all linting errors

PR #330, opened 2015-10-28
NPM 3 compatible version of babelPathResolver

ifyio:
cc/ @matthew-andrews @AlbertoElias

### DIFF
--- a/lib/plugins/babelRuntimePathResolver.js
+++ b/lib/plugins/babelRuntimePathResolver.js
@@ -6,7 +6,9 @@ const path = require('path');
 const falafel = require('falafel');
 
 const prefix = /babel-runtime/;
-const rootDirectory = path.resolve(__dirname, '../../node_modules');
+
+// HACK: Can't require.resolve babel-runtime so use babel-core instead
+const rootDirectory = require.resolve('babel-core').replace(/(.*node_modules).*/, "$1");
 
 module.exports = function babelRuntimePathResolver(file) {
 	this.cacheable = true;
--- a/lib/plugins/babelRuntimePathResolver.js
+++ b/lib/plugins/babelRuntimePathResolver.js
@@ -2,13 +2,12 @@
 // in OBT's node_modules directory and not the module's
 'use strict';
 
-const path = require('path');
 const falafel = require('falafel');
 
 const prefix = /babel-runtime/;
 
 // HACK: Can't require.resolve babel-runtime so use babel-core instead
-const rootDirectory = require.resolve('babel-core').replace(/(.*node_modules).*/, "$1");
+const rootDirectory = require.resolve('babel-core').replace(/(.*node_modules).*/, '$1');
 
 module.exports = function babelRuntimePathResolver(file) {
 	this.cacheable = true;


matthew-andrews:
The problem here is that npm 3 puts `babel-runtime`, `babel-core`, etc under the project level:-

npm 2:-
`node_modules/next-build-tools/node_modules/origami-build-tools/node_modules/babel-runtime`

npm 3:-
`node_modules/babel-runtime`

By using `require.resolve` we don't make any assumptions about the folder hierachy and it just let's NPM deal with figuring out where the modules are.

This probably fixes other cases where babel is required directly by another part of a project.

lib/plugins/babelRuntimePathResolver.js
```diff
@@ -2,11 +2,12 @@
...
 'use strict';
 
-const path = require('path');
 const falafel = require('falafel');
 
 const prefix = /babel-runtime/;
-const rootDirectory = path.resolve(__dirname, '../../node_modules');
+
+// HACK: Can't require.resolve babel-runtime so use babel-core instead
+const rootDirectory = require.resolve('babel-core').replace(/(.*node_modules).*/, '$1');
```
AlbertoElias: what's that replace doing?

matthew-andrews: `require.resolve` gives you the path to the main **file**, so in this something like this:-

`/path/to/my/project/node_modules/babel-core/main.js`

The replace will get rid of everything **after** the last `node_modules`

```js
'/path/to/node_modules/thing/main.js'.replace(/(.*node_modules).*/, '$1');
// => /path/to/thing/node_modules

'/path/to/node_modules/thing/node_modules/other-thing/main.js'.replace(/(.*node_modules).*/, '$1');
// => /path/to/node_modules/thing/node_modules

'/path/to/node_modules/thing/node_modules/other-thing/subdirectory/main-in-a-sub-dir.js'.replace(/(.*node_modules).*/, '$1');
// => /path/to/node_modules/thing/node_modules

```

AlbertoElias: A comment would be very welcomed for non regex experts like me :)

callumlocke: Ah is this in case webpack wants to use the browser main instead of main?

matthew-andrews: No in this case npm has just put the dependency's folders in a difference place, nothing to do with webpack.

matthew-andrews: @AlbertoElias just run those snippets in dev tools and you'll see it works :innocent:

this is what the regex is doing:-
screen shot 2015-10-29 at 07 52 38

callumlocke: @matthew-andrews Oh sorry I read that line out of context. I did a similar fix [here](https://github.com/Financial-Times/origami-build-tools/pull/332/files). I'm just trying to work out what the difference is between yours and mine... Mine doesn't try to establish a 'root', and just resolves every request to where it really is. But yours may be more resiliant in some other way I don't understand :) can you check and say which you think is better

matthew-andrews: Oh your's is very neat.  Simpler than mine :+1:

callumlocke: Thanks but I've realised mine is wrong :) If there's a browser field in the module being required, it will be silently ignored, because the require is rewritten to load the 'main' file directly. It would be better to use [browser-resolve](https://github.com/defunctzombie/node-browser-resolve) instead. But then we're reimplementing logic that webpack does. I think possibly the most sensible way is to use require.resolve but also chop off everything after the package's root (eg the `/lib/index.js` or whatever it is). Then we leave it up to webpack to look at the module's package.json and decide what to do.

matthew-andrews: So is that a :+1: for mine and @ifyio ?  I know ours works … 
…

callumlocke: I'm not sure. What happens if your `rootDirectory` resolves to be the `node_modules` within obt (because user is on npm v2, or just npm v3 with obt linked to a global, say), but then I want to require something that is not within obt but within my module's own `node_modules` dir? Won't it fail in that case?

matthew-andrews: I don't know … :speak_no_evil:  … although that would be broken with the original implementation wouldn't it?

callumlocke: unless I'm mistaken, yes it's already broken in this way. but this brokenness is hidden by the fact that in *most* cases everything will be findable from that same root. the problem comes if you've got something else going on, like maybe you've npm link'd to your own development version of origami-build-tools.

Another edge case would be if you depended on a module foo v1, but origami-build-tools depends on foo v2 for some internal thing (unbeknownst to you) so you end up getting the wrong API and you might get very confusing bugs. So I think we need to specifically rewrite *only* when the requested module is `babel-runtime/**`.

I'll try reworking my one...

callumlocke: oh hang on it already does verify it starts with `babel-runtime`!

callumlocke: if you say yours works then :+1:

matthew-andrews: ok … so is this pr good to merge in your opinion?

callumlocke: yeah sorry about all that faffing

AlbertoElias: Thanks a lot to both of you for taking a good look into this, really appreciate it! :) Merging..

callumlocke:
@AlbertoElias can you release a new patch with this fix?

PR #337, opened 2015-11-04
Resolve babel-runtime on the fly

### LINKED ISSUE #336: npm v3 still doesn't work
Are you sure [this PR](https://github.com/Financial-Times/origami-build-tools/pull/330) actually fixed it? For me I still get this error with the latest master of obt:

```
{ [Error: ModuleNotFoundError: Module not found: Error: Cannot resolve 'file' or 'directory' /Users/callum.locke/code/ft/project-starter-kit/node_modules/babel-runtime/helpers/interop-require-default in /Users/callum.locke/code/ft/project-starter-kit/client/scripts]
  message: 'ModuleNotFoundError: Module not found: Error: Cannot resolve \'file\' or \'directory\' /Users/callum.locke/code/ft/project-starter-kit/node_modules/babel-runtime/helpers/interop-require-default in /Users/callum.locke/code/ft/project-starter-kit/client/scripts',
  showStack: false,
  showProperties: true,
  plugin: 'webpack-stream',
  __safety: { toString: [Function: bound ] } }
```

Using npm 3. @matthew-andrews is it possible you tested this with an npm link'd copy of obt (which would mean a particular directory structure so it wouldn't necessarily work the same)?

callumlocke:
fixes #336

### DIFF
--- a/lib/plugins/babelRuntimePathResolver.js
+++ b/lib/plugins/babelRuntimePathResolver.js
@@ -1,13 +1,10 @@
-// Makes paths to babel-runtime polyfills absolute as they're
-// in OBT's node_modules directory and not the module's
+// Makes paths to babel-runtime polyfills absolute as they might be in OBT's
+// node_modules directory and not the module's, depending on node version
 'use strict';
 
 const falafel = require('falafel');
 
 const prefix = /babel-runtime/;
-
-// HACK: Can't require.resolve babel-runtime so use babel-core instead
-const rootDirectory = require.resolve('babel-core').replace(/(.*node_modules).*/, '$1');
 
 module.exports = function babelRuntimePathResolver(file) {
 	this.cacheable = true;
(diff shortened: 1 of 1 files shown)


AlbertoElias:
@matthew-andrews can you confirm this also fixes it for you?

callumlocke:
a good way to test this 'for real' (ie not testing with a linked local origami-build-tools, which wouldn't have the normal directory structure conditions that you want to test) is to install this branch straight into your product, like this:

```
npm i -D origami-build-tools@financial-times/origami-build-tools#resolve-babel-runtime`
```

then it should go into your node_modules just like a regular module installed from npm, including sharing deps npm3-style.

AlbertoElias:
Shouldn't it also worked when linked?

callumlocke:
Yes it should work either way.

I was just saying that if you test it against an npm link'd copy of obt, it will work fine with or without this fix, so there's not much point in testing that.

PR #339, opened 2015-11-04
Moves demo task to return a stream and fixes not being able to add er…

AlbertoElias:
…ror handlers to build and verify streams

Would be good to double check I haven't left over any mingling tests


lib/tasks/build.js
```diff
@@ -20,14 +20,17 @@ const log = require('../helpers/log');
 module.exports = function(gulp, config) {
 	let jsStream;
 	let sassStream;
+	const buildStream = merge();
 	if (typeof config.watching === 'undefined' || config.watching === 'js') {
-		jsStream = module.exports.js(gulp, config);
+		jsStream = module.exports.js(gulp, config)
+			.on('error', buildStream.emit.bind(buildStream, 'error'));
```
triblondon: Could you maybe self-document this a bit by assigning this to a var?

```
const emitBuildStreamErrorEvent = buildStream.emit.bind(buildStream, 'error');
...
module.exports.js(gulp, config).on('error', emitBuildStreamErrorEvent);
```

lib/tasks/build.js
```diff
@@ -20,14 +20,17 @@ const log = require('../helpers/log');
...
 	}
 	if (typeof config.watching === 'undefined' || config.watching === 'sass') {
-		sassStream = module.exports.sass(gulp, config);
+		sassStream = module.exports.sass(gulp, config)
+			.on('error', buildStream.emit.bind(buildStream, 'error'));
 	}
 	if (jsStream && sassStream) {
-		return merge(jsStream, sassStream);
+		return buildStream.add(jsStream, sassStream);
```
triblondon: If .add() modifies the stream in place and then returns it (I assume that's the case), rather than returning a new stream then it would be better written as:

```
buildStream.add(jsStream, sassStream);
return buildStream;
```

lib/tasks/demo.js
```diff
@@ -21,129 +25,117 @@ const defaultDemoConfig = {
...
-			};
-
-			return build.sass(gulp, sassConfig)
-				.on('error', reject)
-				.on('end', resolve);
-		} else {
-			resolve();
+		const bowerConfig = files.getBowerJson();
+		let prefixSass = '';
+		// If module has o-assets as a dependency, set local demos to use local assets
```
triblondon: This comment (I know it hasn't changed in this refactor) could more helpfully explain why this is the case

lib/tasks/demo.js
```diff
@@ -21,129 +25,117 @@ const defaultDemoConfig = {
...
+			sassPrefix: prefixSass,
+			// For the Sass files to load correctly, they need to be in one of these two directories.
+			// Sass doesn't look in subdirectories and we can't use wildcards
+			sassIncludePaths: ['demos/src', 'demos/src/scss'],
+			sourcemaps: true,
+			buildCss: path.basename(demoConfig.sass).replace('.scss', '.css'),
+			buildFolder: dest
+		};
+
+		return build.sass(gulp, sassConfig);
```
triblondon: You don't appear to be returning anything if `builtFiles.css.indexOf(src) !== -1`

lib/tasks/demo.js
```diff
@@ -21,129 +25,117 @@ const defaultDemoConfig = {
...
+		};
 
-			build.js(gulp, jsConfig)
-				.on('error', reject)
-				.on('end', resolve);
-		} else {
-			resolve();
-		}
-	});
+		return build.js(gulp, jsConfig);
```
triblondon: again

lib/tasks/demo.js
```diff
@@ -216,6 +208,25 @@ function updateOrigamiJsonDemoList(demosBuilt) {
 	}
 }
 
+function errorStream(stream, errorMessage) {
```
triblondon: Comment needed here to explain what you told me when we reviewed it together

lib/tasks/verify.js
```diff
@@ -22,12 +22,17 @@ if (fs.existsSync(excludePath)) {
...
-		module.exports.esLint(gulp, config),
-		module.exports.scssLint(gulp, config),
-		module.exports.lintspaces(gulp, config)
-	);
+	const esLintStream = module.exports.esLint(gulp, config)
+		.on('error', verifyStream.emit.bind(verifyStream, 'error'));
+	const scssLintStream = module.exports.scssLint(gulp, config)
+		.on('error', verifyStream.emit.bind(verifyStream, 'error'));
+	const lintspacesStream = module.exports.lintspaces(gulp, config)
+		.on('error', verifyStream.emit.bind(verifyStream, 'error'));
```
triblondon: Another case where you could just do the bind once.

test/tasks/demo.test.js
```diff
@@ -15,7 +14,7 @@ const oTestPath = 'test/fixtures/o-test';
 const pathSuffix = '-demo';
 const demoTestPath = path.resolve(obtPath, oTestPath + pathSuffix);
 
-describe('Demo task', function() {
+describe('Demo task', () => {
```
triblondon: Generally better to do ES6 syntax refactoring in a separate branch/PR so it doesn't get mixed in

lib/tasks/demo.js
AlbertoElias: default "return undefined" is what I need. I then check if anything got returned

AlbertoElias:
@triblondon I've addressed those issues

triblondon:
https://github.com/Financial-Times/origami-build-tools/pull/339#discussion_r44232072 ?

AlbertoElias:
I'll merge it some point this week

triblondon:
:+1:

AlbertoElias:
@triblondon I've pushed some more changes. This, if all is fine, will be the v5 beta release, and it's what I've been testing the build service with

triblondon:
:+1:

PR #346, opened 2015-11-24
Makes sass stream-combiner an object stream as it was breaking when r…

AlbertoElias:
…unning silent mode test

### DIFF
--- a/lib/tasks/build.js
+++ b/lib/tasks/build.js
@@ -157,7 +157,7 @@ module.exports.sass = function(gulp, config) {
 			remove: config.autoprefixerRemove === undefined ? true : config.autoprefixerRemove
 		};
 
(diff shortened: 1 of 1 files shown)


AlbertoElias:
@JakeChampion this should fix it

JakeChampion:
I pulled down the branch and tested it:
```bash
jake@ft-mw4351 ~/Projects/work/ft/origami/o-share (weback)$ 
npm install -g Financial-Times/origami-build-tools#sass-combine-fix
obt test
Compiling /Users/jake/Projects/work/ft/origami/o-share/main.scss
sass compilation for silent mode: true failed.
```
Not sure if there are any flags I can pass to return more output.

Happy to help tomorrow with this if you are getting stuck

PR #347, opened 2015-11-27
Now adds demo dependencies for js or sass even if the module itself d…

AlbertoElias:
…oes not have any js or sass

I understand it's not the most easiest to read function, so any feedback is very much welcome


AlbertoElias:
@alicebartlett Fixed linting issues

PR #348, opened 2015-11-30
Preserve autoprefixer comments

wheresrhys:
If sass outputs compressed then any autoprefixer comments are stripped out. This change outputs verbose css from sass. As cssclean runs after autoprefixer the css does end up getting minified anyway.

As a small bonus, this will probably make the build slightly quicker


wheresrhys:
yay ta

PR #354, opened 2016-01-04
let babel-runtime be disabled

JakeChampion:
:see_no_evil:


lib/tasks/build.js
```diff
@@ -44,7 +46,7 @@ module.exports.js = function(gulp, config) {
 			// Makes paths to babel-runtime polyfills absolute as they're
 			// in OBT's node_modules directory and not the module's
 			'babel-runtime-path-loader',
-			'babel?optional[]=runtime',
+			babelRuntime ? 'babel?optional[]=runtime' : '',
```
AlbertoElias: you still need to run babel, and you won't need the babel-runtime-path-loader if you're not using babel runtime

JakeChampion:
@AlbertoElias The tests are now passing

PR #357, opened 2016-01-20
Try Node 5

matthew-andrews:


matthew-andrews:
oops

matthew-andrews:
@AlbertoElias can we run the tests on Node 5 too :-)

AlbertoElias:
That will only run obt stuff in node 5. Is that what you mean?

matthew-andrews:
no it should run the test suite in both 4 and 5

AlbertoElias:
What test suite? What I'm saying is that this will only run OBT tests in node 5 in travis, not all origami modules

matthew-andrews:
Yeah… that's fine I think.  We are starting to think about porting our apps to Node 5…  and we wanted to make sure that the build tools work in Node 5.

AlbertoElias:
Ok, fine

Issue #359, opened 2016-01-22
accessibility-- lang attr on html element


StommePoes:
Can we get a lang attribute added to the &lt;html&gt; tag? We figure for now hard-coded as "en" and whenever localisation gets involved it can be changed to a variable. But in any case we want the language to be clearly stated in origami documents and not let things like AT and browsers guess with heuristics and whatnot.
Also example code for developers would want to show this attribute wherever the &lt;html&gt; tag appears.

Going to ping a random a11y-oriented person, @alicebartlett :P  just so one set of those eyes end up here too...

We're still pulling the outer shell of origami for now in our own forks, and it's been explained to me (if I understand correctly) that FT retains control of this? and honestly everyone will want this anyway.

alicebartlett:
Great to hear from you @StommePoes. We'd welcome a PR that added a lang attribute to the html tag. You're right, hardcoded would be fine for now.

StommePoes:
Hi, I've made a change but I don't have permissions. What's a better way to fix?

PR #360, opened 2016-01-28
Add a lang attribute to demos

alicebartlett:
https://github.com/Financial-Times/origami-build-tools/issues/359


alicebartlett:
Next have started using node v5 so that's been added to OBT but there is a failing test for OBT with v5. The failure is not related to this change. @AlbertoElias to fix.

PR #364, opened 2016-02-23
fix error when no JS or Sass streams are returned

onishiweb:
Error occurs when no JS or Sass stream is returned. Check if stream returned is not `undefined` before assigning the event handler.


AlbertoElias:
@onishiweb it seems to be failing ci

AlbertoElias:
I'll merge it, but these test definitely need looking into

PR #371, opened 2016-04-21
Lock down node-sass due to new rule re imports

AlbertoElias:
It's breaking modules, we need time to update them


AlbertoElias:
@alicebartlett it got fixed

PR #373, opened 2016-04-27
Define obt_env JS placeholder and $obt_env SCSS variable, and possibility to define custom vars

roland-vachter:
Define obt_env JS placeholder and $obt_env SCSS variable with the vaue of env in order to be able to make decisions based on the build type.

Also, by adding vars object, obt_env_{varKey} and $obt_env_{varKey} will be passed to the JS and SCSS resources.

### DIFF
--- a/lib/tasks/build.js
+++ b/lib/tasks/build.js
@@ -64,6 +64,7 @@ module.exports.js = function(gulp, config) {
 		].concat(config.loaders || []);
 
 		config.env = config.env || 'development';
+		config.vars = config.vars || {};
 
 		const useSourceMaps = config.sourcemaps || (config.env === 'development');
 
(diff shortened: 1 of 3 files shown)


roland-vachter:
Custom variables are useful for build-time generated fingerprints (which then can be passed to o-assets).

lib/tasks/build.js
```diff
@@ -87,6 +89,17 @@ module.exports.js = function(gulp, config) {
 				sourceMap: useSourceMaps
 			}));
 		}
+
+		plugins.push(new webpack.webpack.DefinePlugin({
+			'obt_env': `"${config.env}"`
```
AlbertoElias: why not create a single object and then push it?

lib/tasks/build.js
```diff
@@ -172,6 +185,15 @@ module.exports.sass = function(gulp, config) {
 
 		const combinedStream = combine.obj(
 			gulp.src(src),
+			inject.prepend(`$obt_env: '${config.env}';`),
```
AlbertoElias: obt comes with a custom gulp plugin called prefixer, you could use that one instead of bringing in a new dependency

roland-vachter:
I'm not actively using this feature anymore, so if you still find it useful, then fine, but I'm not relying on them anymore.

AlbertoElias:
Left a couple of comments. If you could look into those and add tests for it, it'll be good to go

AlbertoElias:
Ok, will close the pr then

PR #375, opened 2016-05-05
add a raw loader for HTML requires

bjfletcher:
This allows the following:

`const template = require('templates/foobar.html')`

where template will become `<h1>Hello world</h1>` if the contents of `templates/foobar.html` is `<h1>Hello world</h1>`.

If people wish to continue and use `requireText`, they can do so. This doesn't break `requireText`. This simply tells `webpack` what to do if it sees a `require('....html')`.


onishiweb:
LGTM 👍

PR #378, opened 2016-06-27
Fix bug with shared demo json

### LINKED ISSUE #368: Demos with the same data but different templates render using the second template
There is something strange going on where 2 demos using the same data (maybe specifically from a file) are rendering using the same template (template given to the second demo in the origami.json). 

I've done some debugging and so far found that before the `buildHtml` function in `demo.js` returns the data for rendering the demos is all present and correct. It's possible then that the combine function at: https://github.com/Financial-Times/origami-build-tools/blob/master/lib/tasks/demo.js#L133 is what's causing the issue, potentially something to do with the double call of `mustache`?

tavvy:
Fixes #368

Looks like the issue was with requiring json and then mutating it


AlbertoElias:
👍

PR #379, opened 2016-06-30
Uses !important in ctm css

AlbertoElias:
Makes it use the same example as the spec: http://origami.ft.com/docs/developer-guide/modules/core-vs-enhanced-experience/#add-styles-to-hide-core-only-html-in-origami


rowanmanning:
LGTM 👍

PR #381, opened 2016-07-07
:zap: improvement: Updates .eslintrc to work in eslint^2

aendrew:
As per the [eslint migration guide](https://github.com/eslint/eslint/blob/master/docs/user-guide/migrating-to-2.0.0.md#language-options), `parserOptions.sourceType` must be set to `module` in order to avoid error: 

> Parsing error: 'import' and 'export' may appear only with 'sourceType: module'

### DIFF
--- a/config/.eslintrc
+++ b/config/.eslintrc
@@ -2,6 +2,9 @@
 	"ecmaFeatures": {
 		"modules": true
 	},
+	"parserOptions": {
+		"sourceType": "module"
+	},
 	"env": {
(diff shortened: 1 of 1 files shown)


AlbertoElias:
Will this also accept commonjs modules, or would ir force the es6 module syntax?

aendrew:
@AlbertoElias Just checked, seems to allow both types.

AlbertoElias:
Cool, thanks!

PR #384, opened 2016-07-25
Support loading of local partials in demos

rowanmanning:
Mustache files in the same directory as the demo template are automatically loaded in and set as partials. This works for files in subdirectories and allows for reuse of HTML in the demos.

This does not support full path resolution, and will not find mustache files in directories further up the tree.

Example:

**`demos/src/example.mustache:`**
```
<p>template</p>
{{>partial1}}
{{>partials/partial2}}
```

**`demos/src/partial2.mustache:`**
```
<p>partial 1</p>
```

**`demos/src/partials/partial2.mustache:`**
```
<p>partial 2</p>
```

**output:**
```
<p>template</p>
<p>partial 1</p>
<p>partial 2</p>
```


AlbertoElias:
I like this, but I think it should have tests and support subdirectories before it's merged

rowanmanning:
Cool, I agree on the tests (of course 😉) and subdirectories. You think it's OK that it's all sync operations?

AlbertoElias:
Most of obt is all sync, so yes

test/helpers/files.test.js
```diff
@@ -130,4 +130,47 @@ describe('Files helper', function() {
...
+		});
+
+		it('returns an array', () => {
+			const mustacheFiles = files.getMustacheFilesList(flatMustacheFiles);
+			expect(mustacheFiles).to.be.an('array');
+		});
+
+		describe('when the directory structure is one level deep', () => {
+
+			it('returns an array of all of the files in the directory', () => {
```
AlbertoElias: of all the mustache files

test/tasks/demo.test.js
```diff
@@ -162,5 +162,27 @@ describe('Demo task', function() {
 
 			demoStream.resume();
 		});
+
+		it('should load local partials', function(done) {
+			fs.mkdirSync('demos/src/partials');
+			fs.writeFileSync('demos/src/test1.mustache', '<div>test1</div>{{>partial1}}', 'utf8');
+			fs.writeFileSync('demos/src/partial1.mustache', '<div>partial1</div>', 'utf8');
+			fs.writeFileSync('demos/src/test2.mustache', '<div>test1</div>{{>partials/partial2}}', 'utf8');
+			fs.writeFileSync('demos/src/partials/partial2.mustache', '<div>partial2</div>', 'utf8');
```
AlbertoElias: If you're not deleting these files afterwards, would it be worth creating them permanently? And they're used by the other test as well

rowanmanning: Yeah can do 👍 I was copying the style of the other tests but I did find it a little odd.

rowanmanning: Actually because all the tests do this, I'd have to refactor the whole thing for it to work. Otherwise the test files I need get overwritten by previous tests.

AlbertoElias: The other tests don't have the templates because those demos are in the origami.json, and the first few tests are very badly written, and require on demo failing because it doesn´t have the templates

AlbertoElias: Might be worth doing then only for the files that aren't used in the other tests?

AlbertoElias:
Might it be worth adding a comment to all the new empty fixture files that explains where they're used? There are quite a few new files

AlbertoElias:
👍 looks good, feel free to merge

PR #385, opened 2016-07-26
Added pa11y test to obt, with its own test. README updated

mariabg:


lib/tasks/test.js
```diff
@@ -264,5 +271,45 @@ module.exports.browserTest = function(gulp, config) {
...
+		},
+		ignore: ['WCAG2AA.Principle2.Guideline2_4.2_4_1.G1,G123,G124.NoSuchID']
+	});
+
+
+	return new Promise(function(resolve, reject) {
+		config = config || {};
+		let src = (config.pa11yPath !== undefined) ? path.join(process.cwd(), config.pa11yPath) : path.join(process.cwd(), '/demos/local/pa11y.html');
+		log.secondary('Running "pa11y test"...');
+		try{
```
AlbertoElias: space

lib/tasks/test.js
```diff
@@ -264,5 +271,45 @@ module.exports.browserTest = function(gulp, config) {
...
 };
 
+module.exports.pa11yTest = function(gulp, config) {
+	const test = pa11y({
+		log: {
+			debug: console.log.bind(console),
+			error: console.error.bind(console),
+			info: console.info.bind(console)
+		},
+		ignore: ['WCAG2AA.Principle2.Guideline2_4.2_4_1.G1,G123,G124.NoSuchID']
```
AlbertoElias: Might be worth commenting what these code reference to, or where to find that information

lib/tasks/test.js
```diff
@@ -264,5 +271,45 @@ module.exports.browserTest = function(gulp, config) {
...
+	return new Promise(function(resolve, reject) {
+		config = config || {};
+		let src = (config.pa11yPath !== undefined) ? path.join(process.cwd(), config.pa11yPath) : path.join(process.cwd(), '/demos/local/pa11y.html');
+		log.secondary('Running "pa11y test"...');
+		try{
+			fs.readFile(src, function(err) {
+				if (err === null) {
+					src = 'file://' + src;
+					test.run(src, function(error, result) {
+						console.log('test', 'callback');
```
AlbertoElias: hanging console.log

lib/tasks/test.js
```diff
@@ -264,5 +271,45 @@ module.exports.browserTest = function(gulp, config) {
...
+						if (error) {
+							console.log('test', error);
```
AlbertoElias: hanging console.log

lib/tasks/test.js
```diff
@@ -5,6 +5,8 @@ const path = require('path');
 const commandLine = require('../helpers/command-line');
 const files = require('../helpers/files');
 const log = require('../helpers/log');
+const pa11y = require('pa11y');
+const cliReporter = require('pa11y/reporter/cli');
```
AlbertoElias: Does this need a separate require?

AlbertoElias: I would also call it pa11yCliReporter if it's going to be global to this module

lib/tasks/test.js
```diff
@@ -264,5 +271,45 @@ module.exports.browserTest = function(gulp, config) {
...
+			error: console.error.bind(console),
+			info: console.info.bind(console)
+		},
+		ignore: ['WCAG2AA.Principle2.Guideline2_4.2_4_1.G1,G123,G124.NoSuchID']
+	});
+
+
+	return new Promise(function(resolve, reject) {
+		config = config || {};
+		let src = (config.pa11yPath !== undefined) ? path.join(process.cwd(), config.pa11yPath) : path.join(process.cwd(), '/demos/local/pa11y.html');
```
AlbertoElias: Probably worth doing this right at the top of the subtask, so you can then check if the file exists, and if not, stop doing anything. Might also be worth not requiring pa11y until this check has been made as requires is what slows down obt the most

lib/tasks/test.js
```diff
@@ -264,5 +271,45 @@ module.exports.browserTest = function(gulp, config) {
...
+							resolve(result);
+						}
+					});
+				} else {
+					log.primary(`Could not open ${src}`);
+					resolve(err);
+				};
+			});
+		}catch(err){
+			console.log('error', err);
```
AlbertoElias: This should probably be a log.secondaryError. We use the log helper to colour console.logs

lib/tasks/test.js
```diff
@@ -264,5 +271,45 @@ module.exports.browserTest = function(gulp, config) {
...
+							cliReporter.results(result, src);
+							resolve(result);
+						}
+					});
+				} else {
+					log.primary(`Could not open ${src}`);
+					resolve(err);
+				};
+			});
+		}catch(err){
```
AlbertoElias: spaces

test/tasks/test.test.js
```diff
@@ -0,0 +1,68 @@
...
+
+		before(function(done) {
+			fs.copySync(path.resolve(obtPath, oTestPath), testTestPath);
+			process.chdir(testTestPath);
+			const demoStream = demo(gulp, {
+				demoConfig: 'origami.json',
+				demoFilter: ['pa11y']
+			});
+			demoStream
+			.on('end', done);
```
AlbertoElias: This can probably go on the previous line

test/tasks/test.test.js
```diff
@@ -0,0 +1,68 @@
...
+
+		after(function() {
+			process.chdir(obtPath);
+			fs.removeSync(testTestPath);
+		});
+
+		it('should not fail when the file is not found', function(done){
+			// run pa11y subtask
+			let res = test.pa11yTest(gulp, {
+				// set an empty path
```
AlbertoElias: an invalid path?

test/tasks/test.test.js
```diff
@@ -0,0 +1,68 @@
...
+				pa11yPath: './file.html'
+			});
+			res
+			.then(function() {
```
AlbertoElias: Could you indent this for clarity?

test/tasks/test.test.js
```diff
@@ -0,0 +1,68 @@
...
+				expect(true).to.be(false);
```
AlbertoElias: Maybe a comment here saying that it should throw an error, not resolve

test/tasks/test.test.js
```diff
@@ -0,0 +1,68 @@
...
+				done();
+			})
+			.catch(function errorHandler(err) {
+				expect(err);
```
AlbertoElias: it should expect the error to be something or contain a specific string

test/tasks/test.test.js
```diff
@@ -0,0 +1,68 @@
...
+				done();
+			});
+		});
+
+		it('should run pa11y correctly', function(done) {
+			// run pa11y subtask
+			let res = test.pa11yTest();
```
AlbertoElias: const

test/tasks/test.test.js
```diff
@@ -0,0 +1,68 @@
...
+			res
+			.then(function() {
+				done();
```
AlbertoElias: doesn't the subtask resolve with the results? you could expect the result to be the expected one

test/tasks/test.test.js
```diff
@@ -0,0 +1,68 @@
...
+			.on('end', done);
+			demoStream.resume();
+		});
+
+		after(function() {
+			process.chdir(obtPath);
+			fs.removeSync(testTestPath);
+		});
+
+		it('should not fail when the file is not found', function(done){
```
AlbertoElias: You're saying it should not fail, but your assertions are on the promise catch

AlbertoElias:
Great job on this! I've left a few comments, some are style related.  Let me know if anything isn't clear

README.md
```diff
@@ -166,6 +166,9 @@ Config accepts:
 	- environments: `String` Comma separated list of environments from your nightwatch config file to run your tests on. (Default: `chrome_latest,chrome_latest-1,firefox_latest,firefox_latest-1,ie8_Grid,ie9_Grid,ie10_Grid,ie11_Grid,safari7_Grid`)
 	- testsPath: `String` Relative path from your project's root directory to where your nightwatch tests are. (Default: `test/browser/tests`)
 	- browserTestFailBuild: `Boolean` by default the failed browser tests will not fail the task, however you can override the default behaviour by setting this option to `true`. Default: `false`
+* __pa11yTest(gulp, config)__ The test will not fail if the pa11y.html file is not find. Config accepts:
```
rowanmanning: Maybe we should document that PhantomJS is required for this to run?

lib/tasks/test.js
```diff
@@ -264,5 +270,46 @@ module.exports.browserTest = function(gulp, config) {
...
+				const test = pa11y({
+					log: {
+						debug: console.log.bind(console),
+						error: console.error.bind(console),
+						info: console.info.bind(console)
+					},
+					ignore: ignore
+				});
+
+				src = 'file://' + src;
```
rowanmanning: I think this might restrict us at some point. Could we only add "file://" if a protocol isn't already set? E.g.

```
foo/bar          becomes  file://foo/bar
file://foo/bar   becomes  file://foo/bar
http://example/  becomes  http://example/

**Comment by rowanmanning** [comment]
Looking great 👍 I've added a couple of comments but hopefully nothing major

File: README.md (hunk 17 above)
Comment (AlbertoElias) [inline README.md]: Is OBT installing it via some dependency then? Because if not, I don't understand why the pa11y subtask would work in ci

Comment (rowanmanning) [inline README.md]: No, Travis always has a version of PhantomJS installed :)

File: lib/tasks/test.js [hunk 19]
```diff
@@ -6,6 +6,7 @@ const commandLine = require('../helpers/command-line');
 const files = require('../helpers/files');
 const log = require('../helpers/log');
 
+
```
AlbertoElias: unnecessary new line

lib/tasks/test.js
```diff
@@ -136,6 +141,7 @@ function pollHeroku(appUrl) {
 		});
 }
 
+
```
AlbertoElias: new line again

lib/tasks/test.js
```diff
@@ -264,5 +270,46 @@ module.exports.browserTest = function(gulp, config) {
...
+				test.run(src, function(error, result) {
+					if (error) {
+						reject(error);
+					} else {
+						pa11yCliReporter.results(result, src);
+						resolve(result);
+					}
+				});
+			} else {
+				log.primary(`Could not open ${src}`);
```
AlbertoElias: if the "running pa11y test" is a `log.secondary`, maybe this should be one too?

lib/tasks/test.js
```diff
@@ -264,5 +270,46 @@ module.exports.browserTest = function(gulp, config) {
...
+				const test = pa11y({
+					log: {
+						debug: console.log.bind(console),
+						error: console.error.bind(console),
+						info: console.info.bind(console)
+					},
+					ignore: ignore
+				});
+
+				src = (src.indexOf('file://') === -1 && src.indexOf('http://') === -1 && src.indexOf('https://') === -1 ) ? 'file://' + src : src;
```
AlbertoElias: probably worth separating all the checks into separate lines so this line isn't that long to make it more readable

test/tasks/test.test.js
```diff
@@ -0,0 +1,88 @@
...
+			// run pa11y subtask
+			const res = test.pa11yTest(gulp, {
+				// set an invalid path
+				pa11yPath: './file.html'
+			});
+			res
+				.then(function(results) {
+					// it should throw an error, not resolve
+					//expect(results.errno).to.be(-2);
```
AlbertoElias: is this comment still necessary?

test/tasks/test.test.js
```diff
@@ -0,0 +1,88 @@
...
+
+		it('should not fail when the file is not found', function(done){
+			// run pa11y subtask
+			const res = test.pa11yTest(gulp, {
+				// set an invalid path
+				pa11yPath: './file.html'
+			});
+			res
+				.then(function(results) {
+					// it should throw an error, not resolve
```
AlbertoElias: why should it throw an error?

test/tasks/test.test.js
```diff
@@ -0,0 +1,88 @@
...
+		it('should run pa11y correctly', function(done) {
+			// run pa11y subtask
+			const res = test.pa11yTest();
+			res
+			.then(function(results) {
+				expect(results[0].type).to.be('error');
+				done();
+			})
+			.catch(function errorHandler() {
+				// it should throw an error, not resolve
```
AlbertoElias: it should be running correctly, so it should be resolving

test/tasks/test.test.js
```diff
@@ -0,0 +1,88 @@
...
+				// set custom ignore options
+				pa11yIgnore: 'error;notice'
+			});
+			res
+			.then(function(results) {
+				expect(results[0].type).to.be('warning');
+				done();
+			})
+			.catch(function errorHandler() {
+				// it should throw an error, not resolve
```
AlbertoElias: same as above

AlbertoElias:
Left a few nitpicky comments. If you can fix that, and clean up the commit history and this will be good to go! :D

lib/tasks/test.js
JakeChampion: You can use the url module built into node for this -- https://nodejs.org/dist/latest-v6.x/docs/api/url.html

PR #386, opened 2016-08-16
Deprecate the expanded property for demos

onishiweb:
Add a failing test for the verify task which looks for the expanded property in component demos.

Do not merge until the new Registry has been deployed.


onishiweb:
@AlbertoElias this is ready to review now I've updated the tests. I think the origami.json tests could probably be split up more like you said before but this is a good starting point.

AlbertoElias:
👍

Issue #390, opened 2016-09-05
Analytics


JakeChampion:
## Problem

We can not tell how many:
- installations we have
- how many users we have
- how often each command is used
- which versions of Node are they installed/executed under

This information is key in prioritising what work should be done and how it should be done.
## Proposed Solution

Add analytics for each of these stats and possibly more, using the same set-up as our services.
Graphite/Grafana
This solution will of course not work for users when they are offline, I think that is okay as most of our users are online when they are using OBT.
We can decide on whether analytics should be opt-in or opt-out.

Limess:
The metrics added in https://github.com/Financial-Times/origami-build-tools/pull/394 seem to have been removed as part of https://github.com/Financial-Times/origami-build-tools/pull/417.

Was that intentional @JakeChampion? Is this still relevant?

PR #394, opened 2016-09-07
Add feature: Analytics

JakeChampion:
This makes a start to addressing #390.

This adds Graphite reporting capabilities and uses them to track the major version of node being used when executing OBT whether from the command line or programatically.


rowanmanning:
This looks great 👍

PR #396, opened 2016-09-08
close the socket once a write is completed

JakeChampion:
Without this, the node process would not close until the tcp connection automatically closes itself.


rowanmanning:
Looks good 👍 were you missing some data?

rowanmanning:
Looks even better now 😄

PR #397, opened 2016-09-08
Let node 5 fail on CI

JakeChampion:
Node 5 is no longer supported by Node, we should also no longer support Node 5 and concentrate on Node 4 and Node 6 until October.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ sudo: false
 language: node_js
 node_js:
   - "4"
+  - "5"
   - "6"
 script:
   - gulp install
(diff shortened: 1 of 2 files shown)


rowanmanning:
Would it be better to [mark the node 5 tests as allowed to fail](https://docs.travis-ci.com/user/customizing-the-build#Rows-that-are-Allowed-to-Fail)? I think it's still being used in Node 5 in places and it might be useful to be able to quickly tell whether the build still passes

JakeChampion:
That's a good idea :-)

rowanmanning:
Looks great 👍

PR #427, opened 2017-04-11
Bring Karma and Browserstack into OBT

JakeChampion:


lib/tasks/test.js
```diff
@@ -176,5 +176,29 @@ const karmaTest = function (gulp, config) {
...
+	const bower = files.getBowerJson();
+
+	if (bower) {
+		return globby(['bower_components/*/origami.json', 'origami.json'])
+			.then(dependencies => {
+				const requiredFeatures = [];
+
+				for (const dependency of dependencies) {
+					const origami = require(path.resolve(dependency));
+					const { browserFeatures: { required: features = [] } = {} } = origami;
```
rowanmanning: I don't know how I feel about this, it's not _immediately_ readable but I'm not as familiar with destructuring as you

PR #438, opened 2017-04-20
obt test -- start using listr to display the cli results

JakeChampion:


package.json
```diff
@@ -52,6 +52,7 @@
     "karma-sinon": "^1.0.5",
     "karma-sourcemap-loader": "^0.3.7",
     "karma-webpack": "^2.0.3",
+    "listr": "^0.11.0",
     "meow": "^3.7.0",
     "merge-stream": "^1.0.0",
     "minimist": "^1.1.1",
```
onishiweb: Should `globby` also be saved in the deps?

JakeChampion: It already is in there :-) -- https://github.com/Financial-Times/origami-build-tools/blob/81411503e609df679f5016775bfe68ea1aac21c9/package.json#L30

onishiweb: Whoops,  I missed that :/

PR #447, opened 2017-05-16
Move install tasks to separate files and remove user customisation options

JakeChampion:


test/tasks/insta..test.js
```diff
@@ -0,0 +1,67 @@
+/* eslint-env mocha */
```
rowanmanning: File name typo? Should be `install.test.js` maybe?

PR #458, opened 2017-05-31
Makes files module and all consumers async

JakeChampion:


lib/tasks/sass.js
```diff
@@ -2,65 +2,66 @@
...
-const nonSilentCompilation = function() {
+const nonSilentCompilation = function () {
 	return silentCompilationTest(false);
 };
 
 function silentCompilationTest(silent) {
 	const sass = require('./build-sass');
 	const silentSass = require('../plugins/silent-sass');
 	const nodeSass = require('node-sass');
+	return files.getMainSassPath()
```
rowanmanning: This stuff here is quite difficult to read, Is it possible for the `then`s to be chained in a more flat way?

JakeChampion: Definitely. I first wrote this using `async` functions and then converted to promises. I'll do this after lunch.

PR #473, opened 2017-07-05
use raw-loader for .txt files

JakeChampion:
n-webpack uses handlebars to compile .html files and obt uses the raw-loader to return the html as a string. This caused issues on o-comments as handlebars returns a function instead of a string.

To avoid this confusion, we have both agreed to use .txt to signify require calls which should return strings of the file being required.

This PR and a PR to change o-comment-ui to use .txt instaed of .html will mean that next-article can be updated to the new origami components! 🎉


JakeChampion:
Related to this -> https://github.com/Financial-Times/n-ui/pull/1011

PR #479, opened 2017-08-31
Allow remote demo data

notlee:
In some cases we would like to use external data when building component demos so our demos match reality. For example in `o-footer`:
https://github.com/Financial-Times/o-footer/issues/81

For reliability of build this should be used sparingly. The `o-footer` example would depend on the navigation service, a "platinum" application.

E.g `origami.json` `demos` config using a remote data source (new ability):
```
{
    "name": "footer",
    "title": "Dark theme",
    "template": "demos/src/footer-dark.mustache",
    "data": "https://www.ft.com/__origami/service/navigation/v2/menus/footer/?source=test",
    "description": "This is the standard masterbrand footer used on most pages."
}
```

E.g `origami.json` `demos` config using a local data source (current):
```
{
    "name": "footer",
    "title": "Dark theme",
    "template": "demos/src/footer-dark.mustache",
    "data": "demos/src/footer.json",
    "description": "This is the standard masterbrand footer used on most pages."
}
```

### DIFF
--- a/lib/tasks/demo-build.js
+++ b/lib/tasks/demo-build.js
@@ -90,6 +90,7 @@ function buildJs(buildConfig) {
 }
 
 function loadRemoteDemoData(url) {
+	const genericErrorMessage = 'Could not load remote demo data.';
 	return request.get({
 		uri: url,
 		json: true
@@ -98,7 +99,15 @@ function loadRemoteDemoData(url) {
 		if (typeof data === 'object') {
 			return data;
 		}
-		const e = new Error(`${url} did not provide valid JSON.`);
+		throw new Error(`${genericErrorMessage} ${url} did not provide valid JSON.`);
+	}).catch((err) => {
+		if (err.name === 'RequestError') {
+			err.message = `${genericErrorMessage} ${url} does not appear to be valid.`;
+		}
+		if (err.name === 'StatusCodeError') {
+			  err.message = `${genericErrorMessage} ${url} returned a ${err.statusCode} status code.`;
+		}
+		const e = new Error(err.message);
 		e.stack = '';
 		throw e;
 	});
--- a/lib/tasks/demo-build.js
+++ b/lib/tasks/demo-build.js
@@ -125,7 +125,7 @@ function loadLocalDemoData(dataPath) {
 function loadDemoData(buildConfig) {
 	if (typeof buildConfig.demo.data === 'string') {
 		const dataPathIsAbsolute = /^(?:https?:)\/\//.test(buildConfig.demo.data);
-		const dataPath = (dataPathIsAbsolute ? buildConfig.demo.data : path.join(buildConfig.cwd, '/' + buildConfig.demo.data));
+		const dataPath = dataPathIsAbsolute ? buildConfig.demo.data : path.join(buildConfig.cwd, '/' + buildConfig.demo.data);
 		if (dataPathIsAbsolute) {
 			return loadExternalDemoData(dataPath);
 		} else {
--- a/lib/tasks/demo-build.js
+++ b/lib/tasks/demo-build.js
@@ -99,17 +99,19 @@ function loadRemoteDemoData(url) {
 		if (typeof data === 'object') {
 			return data;
 		}
-		throw new Error(`${genericErrorMessage} ${url} did not provide valid JSON.`);
+		const err = new Error(`${genericErrorMessage} ${url} did not provide valid JSON.`);
+		err.stack = '';
+		throw err;
 	}).catch((err) => {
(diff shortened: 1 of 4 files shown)


lib/tasks/demo-build.js
```diff
@@ -88,29 +89,48 @@ function buildJs(buildConfig) {
...
-								}
-							} catch (error) {
-								const e = new Error(`${dataPath} is not valid JSON.`);
-								e.stack = '';
-								throw e;
-							}
-						});
-				}
-			});
+		const dataPathIsAbsolute = /^(?:https?:)\/\//.test(buildConfig.demo.data);
```
JakeChampion: Can we change the variable name from `dataPathIsAbsolute` to `dataPathIsHTTP` or `dataPathIsRemote`? Absolute makes me think that an absolute file-system path would set this variable to true (which it doesn't).

lib/tasks/demo-build.js
```diff
@@ -88,29 +89,48 @@ function buildJs(buildConfig) {
 	}
 }
 
+function loadExternalDemoData(url) {
+	return request.get({
```
JakeChampion: What happens in the different failure scenarios?
- URL is not valid
- Response returns a HTTP status code equal or greater than 400
- Response returns a redirect
- Response is not valid JSON

notlee: URL is not valid: Error thrown with stack trace. Do you think it aught to be caught and the stacktrace removed as with file not found?
screen shot 2017-09-11 at 13 48 22

Response returns a HTTP status code equal or greater than 400: Error thrown with stack trace.
screen shot 2017-09-11 at 13 51 43

Response returns a redirect: `request-promise-native` follows up to 10 redirects by default.

Response is not valid JSON: I've now caught this case, introducing a more helpful error.
screen shot 2017-09-11 at 13 34 06

lib/tasks/demo-build.js
```diff
@@ -90,6 +90,7 @@ function buildJs(buildConfig) {
 }
 
 function loadRemoteDemoData(url) {
+	const genericErrorMessage = 'Could not load remote demo data.';
```
JakeChampion: Could we include the url in the error message please, that way the user might be able to understand why we could not load remote demo data

JakeChampion: Ignore this, it looks like we add it later on 👌

lib/tasks/demo-build.js
```diff
@@ -88,29 +89,62 @@ function buildJs(buildConfig) {
...
+		throw new Error(`${genericErrorMessage} ${url} did not provide valid JSON.`);
+	}).catch((err) => {
+		if (err.name === 'RequestError') {
+			err.message = `${genericErrorMessage} ${url} does not appear to be valid.`;
+		}
+		if (err.name === 'StatusCodeError') {
+			  err.message = `${genericErrorMessage} ${url} returned a ${err.statusCode} status code.`;
+		}
+		const e = new Error(err.message);
+		e.stack = '';
```
rowanmanning: Why do we remove the stack from the error here? Could be useful in debugging

lib/tasks/demo-build.js
```diff
@@ -88,29 +89,62 @@ function buildJs(buildConfig) {
...
+		throw e;
+	});
+}
+
+function loadLocalDemoData(dataPath) {
+	return fileExists(dataPath)
+		.then(exists => {
+			if (exists) {
```
rowanmanning: Do we want to throw an error if the local file doesn't exist? I know it's not in there currently but seems like it should if we're modifying this code anyway.

lib/tasks/demo-build.js
notlee: Good question. Me and @JakeChampion chatted about that for a while. I think we opted for a more concise error to make OBT more friendly `Could not load remote demo data. ${remoteDataUrl} returned a ${statusCode} status code.` is straight to the point, a stack trace of vendor code is not so helpful. Perhaps the best-of-both would be to remove the stack trace only for potentially common cases we handle `RequestError` `StatusCodeError`.

rowanmanning: But with a new `Error` thrown (the one you create) it doesn't include any vendor code in the stack trace anyway right?

Just feels like this library code maybe isn't the place to decide whether the stack trace is useful to a dev who's trying to debug later 🙂. If we're stripping stack traces from certain errors to make the output more friendly, should that be handled by the [command line part of OBT](https://github.com/Financial-Times/origami-build-tools/blob/major/lib/origami-build-tools-cli.js#L115)?

notlee:
@rowanmanning I made a couple of tweaks that should address your comments.

PR #510, opened 2018-02-13
Require demo titles in origami.json.

notlee:
Inline with the Origami spec and registry expectations.

All actively maintained Origami maintained `isOrigami` components should be covered when the PRs created today are merged. Deprecated or non-origami-maintained components will get caught by this new `obt verify` requirement.

Does this effect the build service?

### DIFF
--- a/lib/tasks/verify-origami-json.js
+++ b/lib/tasks/verify-origami-json.js
@@ -45,11 +45,11 @@ function origamiJson(config) {
 								result.push('The expanded property has been deprecated. Use the "hidden" property when a demo should not appear in the Registry.');
 							}
 
-							const hasNoTitle = componentDemos.some(function (demo) {
-								return ! Object.prototype.hasOwnProperty.call(demo, 'title');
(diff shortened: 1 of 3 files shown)


test/unit/fixtures/o-test/origami.json
```diff
@@ -11,6 +11,7 @@
     },
     "demos": [
         {
+              "title": "Test module",
```
JakeChampion: Need to add a test case where title is an empty string.

lib/tasks/verify-origami-json.js
```diff
@@ -44,6 +44,13 @@ function origamiJson(config) {
 							if (hasExpanded) {
 								result.push('The expanded property has been deprecated. Use the "hidden" property when a demo should not appear in the Registry.');
 							}
+
+							const hasNoTitle = componentDemos.some(function (demo) {
+								return ! Object.prototype.hasOwnProperty.call(demo, 'title');
```
JakeChampion: This checks for existence of a property but not whether that property is a string nor whether it is an empty string.

PR #520, opened 2018-04-19
Make obt test work with npx

JakeChampion:
`npx origami-build-tools test` currently fails because `execa` can not find `mocha`. This is because `npx` installs modules into a directory that `execa` doesn't look into. We can tell `execa` where to look using the `localDir` option.


gvonkoss:
🕵️

Issue #508, opened 2018-02-07
Add line number to compilation errors


gvonkoss:
Is this possible? Occasionally when I'm compiling, I'll get something like: 
```
✖ Compiling Sass
    → Undefined variable: "$_some-variable".
```
or
```
  ✖ Compiling Sass
    → File to import not found or unreadable: src/scss/variables.
    → Parent style sheet: stdin
```
Which is all well and good with small codebases, but maybe not if things start getting bigger?

PR #527, opened 2018-05-16
Reformat SASS build error to include file, line, and column.

notlee:
Before and after:

screen shot 2018-05-16 at 15 39 00


notlee:
Addresses: https://github.com/Financial-Times/origami-build-tools/issues/508

PR #534, opened 2018-07-24
First pass at adding package.json generation to obt

aendrew:
Please see forthcoming comment on Financial-Times/ft-origami#481.

N.b., there are integration tests on this but not unit; if this is actually functionality desirable in `obt` I'll happily write them and amend this PR.


aendrew:
Not sure why Travis is failing; seems like a Whitesource violation. I haven't added any deps so pretty sure it's not my PR causing it.

PR #536, opened 2018-07-27
boilerplate

gvonkoss:
yay boilerplate 🎉 
This builds the whole `o-component-boilerplate` folder structure into the directory that the command is called.

It accepts a component name and populates all of the files in the tree, and will not run if the parent folder already exists.

I have not yet added anything relating to branding, because it is still in experimental stages.

I have not changed much about the current boilerplate, so this may be a good opportunity to do so. If not, then we can iterate on this in the future.

Thoughts?


lib/tasks/boilerplate-tree.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+
+const fs = require('fs-extra');
+const path = require('path');
+const template = require('./../../templates/boilerplate.js');
+
+const buildTree = (root, name) => {
+	const generate = (file, template) => fs.outputFile(path.join(root, file), template);
+
+	return Promise.all([
```
notlee: This use of `Promise.all`, beautiful 👏

test/integration/boilerplate/boilerplate.test.js
```diff
@@ -0,0 +1,91 @@
...
+					})
+					.then(() => {
+						return fileExists(`${defaultName}/src/js/${defaultName}.js`);
+					})
+					.then(exists => {
+						proclaim.ok(exists);
+					});
+			});
+
+			it('should error if component folder exists', function () {
```
notlee: Duplicate "should error if component folder exists". Extra safe 💪

templates/component-boilerplate/main-scss.js
```diff
@@ -0,0 +1,13 @@
+'use strict';
+
+module.exports = (name) => {
+	return `@import 'scss/variables';
+@import 'scss/mixins';
+
+@if ($${name.original}-is-silent == false) {
+@include ${name.camelCase}();
```
notlee: Idententation is off here within the if statement.

templates/component-boilerplate/main-js.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+module.exports = (name) => {
+	return `import ${name.camelCase} from './src/js/${name.original}';
+
+const constructAll = function() {
```
notlee: Indentation within the function

notlee:
This is the greatest -- we should have had it sooner! :D

Just spotted a coupld of things:
- The demo mustache file is missing so `obt build` fails.
- `functions.scss` would be great too.
-  Other nitpicks as above.

PR #551, opened 2018-08-31
No longer run snyk via make whitesource or make ci

JakeChampion:
Synk is run via the GitHub integration and has it's own specific configuration, keeping two copies of configuration for Snyk (one in Snyk's UI and one in this repo) is leading to them becoming out of sync and causing builds to needlessly fail.

I've gone with removing the configuration within the repo as it is the one which is out of sync, I am happy to remove the other one if we think that is the better option to take.


notlee:
Whoops. I've duplicated this here: https://github.com/Financial-Times/origami-build-tools/pull/566

PR #554, opened 2018-09-12
Added TODO-comment-finding eslint rule

carlesandres:
The added rule will report any `//TODO`, `//FIXME` and `//xxx` code comments as linting warnings.

It will make and OBT-dependant project that uses OBT's eslint rules (e.g., `o-ads` is doing this now), to report those to-do comments.

Interesting fact: linting OBT itself is currently reporting 5 of those comments.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -17,6 +17,7 @@ const origamiServicesConfig = merge({},
 		"rules": {
 			"no-console": "off",
 			"no-await-in-loop": "off",
+			"no-warning-comments": "warn",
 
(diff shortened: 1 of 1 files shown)


.eslintrc.js
```diff
@@ -17,6 +17,7 @@ const origamiServicesConfig = merge({},
 		"rules": {
 			"no-console": "off",
 			"no-await-in-loop": "off",
+			"no-warning-comments": "warn",
```
JakeChampion: This should be added to the [component eslint config](https://github.com/Financial-Times/origami-build-tools/blob/master/config/.eslintrc.js) instead of this eslint config. This eslint config is only for the OBT codebase.

carlesandres: Makes sense. Closing.

carlesandres:
Closing as this is more suited at the component-level

PR #561, opened 2018-10-17
Fix CircleCI badge

alicebartlett:


notlee:
Thanks for flagging and opening a PR. 👍 Turns out it was a different issue with our circleci config and the badge is working now, so closing this PR.

PR #562, opened 2018-10-23
Brand param

notlee:
**Add --brand option**
To make it easier to develop branded components, a `brand` flag prepends the brand SCSS variable when building.

E.g. to build demos for the internal brand:
`obt demo --run-server --watch --brand=internal`

**Test brand compilation**
As well as testing silent mode, OBT now also tests compilation for each brand the component supports, as defined in the components `origami.json` file.
screen shot 2018-10-22 at 18 25 35


lib/helpers/files.js
```diff
@@ -121,6 +126,16 @@ function getModuleName(cwd) {
 		});
 }
 
+function getModuleBrands(cwd) {
+	return getOrigamiJson(cwd)
+		.then(origamiJson => {
+			if (origamiJson && origamiJson.brands) {
```
rowanmanning: Just to be safe, should we change to this?
```suggestion
			if (origamiJson && origamiJson.brands && Array.isArray(origamiJson.brands)) {
```

PR #566, opened 2018-11-21
Review Snyk notices.

notlee:
- Upgrades to a new major of chokidar.
- Bumps `karma-webpack`, `bower-resolve-webpack-plugin`, `babel-loader`.
- Ignores other notices which appear not to affect production systems. Upgrading major dependancies is in our roadmap, which will clear these up.


notlee:
Also no longer runs whitesource. Duplicates https://github.com/Financial-Times/origami-build-tools/pull/551

PR #573, opened 2018-12-13
Change Karma's debug.html to the mocha web reporter so we can see the test results in page instead of in the console

JakeChampion:
Currently when debugging tests you need to:
1. Run Karma
2. Open http://localhost:9876/debug.html
3. Open console

This PR removes step 3 :-)

### DIFF
--- a/config/karma.config.js
+++ b/config/karma.config.js
@@ -25,7 +25,11 @@ module.exports.getBaseKarmaConfig = function() {
 
 			client: {
 				// Capture all console output and pipe it to the terminal.
-				captureConsole: false
+				captureConsole: false,
+				mocha: {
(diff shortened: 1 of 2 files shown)


notlee:
Looks great for pure JS tests, but I'm not sure for components which test aspects of the DOM. 

As the tests run, the sandbox is pushed incrementally further down which might be a nuisance when stepping through tests to debug issues with the DOM/CSS.

I wonder if adding the test results to the page might interfere with the tests themselves too. It appears there are failures for `o-table`, but they may be poor tests.

Before (o-table, passing)
screenshot 2018-12-13 at 16 54 10

After (o-table, failures)
screenshot 2018-12-13 at 16 42 59

JakeChampion:
This looks to cause issues for components which have tests that use the DOM, which is a shame.

PR #574, opened 2018-12-14
Let's use npm ci instead of npm i

JakeChampion:
The master branch had a package-lock.json which was out-of-sync with package.json, we can ensure that never happens by using `npm ci` in CircleCI instead of `npm i`. Using `npm ci` has an added benefit of simplifying the CircleCI config because it guarantees that the node_modules folder will be the same.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,36 +24,6 @@ workflows:
               only: /^v.*/
             branches:
               ignore: /.*/
-references:
-  restore_npm_cache: &restore_npm_cache
-    restore_cache:
-      keys:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -24,50 +24,16 @@ workflows:
...
       - image: circleci/node:latest-browsers
     steps:
-      - *versions
       - checkout
-      - *restore_npm_cache
-      - *install
-      - *test
-      - *save_cache_local
-      - *save_cache_global
+      - run: npm ci || npm i
```
JakeChampion: We fallback to `npm i` if the version of npm doesn't suppport `npm ci`. This is only required for node 6.

PR #580, opened 2019-01-23
Don't look at me

JakeChampion:


lib/tasks/test-sass.js
```diff
@@ -16,11 +16,12 @@ const trueTest = function (config) {
 		const componentBase = path.join(__dirname, '../../');
 		const sassFile = path.join(__dirname, 'index.test.scss');
 		const sassTrue = require('${require.resolve('sass-true')}');
+		const getSassIncludePaths = ${files.getSassIncludePaths.toString()};
```
JakeChampion: This is wild

JakeChampion:
Currently this will not build demos correctly for components which use o-assets --> https://github.com/Financial-Times/origami-build-tools/blob/master/lib/tasks/demo-build.js#L37

This can be a future improvement 👍

JakeChampion:
Do we need to update this file also? https://github.com/Financial-Times/origami-build-tools/blob/master/templates/component-boilerplate/ci-config-yml.js

JakeChampion:
This might need to be updated to get testing to work --> https://github.com/Financial-Times/origami-build-tools/blob/cf219aa5029754feb0f98c63db87f2af4a439786/lib/helpers/construct-polyfill-url.js#L9

PR #587, opened 2019-02-26
✨Kate makes updates ✨

sbinlondon:
Rowan and I went through and looked at the differences between the new o-message boilerplate and the boilerplate generated for my imagined component, o-mopsa. Based on differences I saw, Rowan suggested some changes to the boilerplate.


templates/component-boilerplate/ci-config-yml.js
```diff
@@ -5,37 +5,36 @@ module.exports = () => {
...
+      - run: git clean -fxd
+      - run: npx occ 0.0.0
+      - run: $HOME/.local/bin/obt install --ignore-bower
+      - run: $HOME/.local/bin/obt test --ignore-bower
+    publish_to_npm:
+      docker:
+        - image: circleci/node:10
+      steps:
+        - checkout
+        - run: npx occ ${CIRCLE_TAG##v}
```
JakeChampion: It is getting turned into a template string interpolation, to avoid that we need to escape it like this:
```suggestion
        - run: npx occ \$\{CIRCLE_TAG##v}
```

lib/tasks/boilerplate.js
```diff
@@ -22,7 +22,8 @@ const buildTree = (root, name) => {
 		generate('/demos/src/demo.mustache', template.demoMustache(name)),
 		generate('/demos/src/pa11y.mustache', template.demoPa11y(name)),
 		generate('/demos/src/demo.scss', template.demoSass(name)),
-		generate(`/src/js/${name.original}.js`, template.srcJs(name)),
+		generate(`/src/js/${name.plainName}.js`, template.srcJs(name)),
+		generate('/src/scss/_brand.scss', ''),
```
notlee: A common pattern for branded components is to wrap `oBrandGet` and `oBrandSupportsVariant` in a private, component specific mixin `_oComponentGet`, and`_oComponentSupports`. Maybe it's worth adding that to the boilerplate?

Here's some examples:
- https://github.com/Financial-Times/o-table/blob/master/src/scss/_brand.scss#L3
- https://github.com/Financial-Times/o-overlay/blob/master/src/scss/_brand.scss#L3
- https://github.com/Financial-Times/o-buttons/blob/master/scss/_brand.scss#L3

templates/component-boilerplate/read-me.js
```diff
@@ -88,7 +84,7 @@ State | Major Version | Last Minor Release | Migration guide |
 
 If you have any questions or comments about this component, or need help using it, please either [raise an issue](https://github.com/Financial-Times/${name.original}/issues), visit [#ft-origami](https://financialtimes.slack.com/messages/ft-origami/) or email [Origami Support](mailto:origami-support@ft.com).
 
-## Licence
+## License
```
JakeChampion:
```suggestion
## Licence
```

templates/component-boilerplate/ci-config-yml.js
sbinlondon: 🌮 (I was going to do 🎉 but github suggested taco first so...)

templates/component-boilerplate/read-me.js
```diff
@@ -88,7 +84,7 @@ State | Major Version | Last Minor Release | Migration guide |
...
 
-This software is published by the Financial Times under the [MIT licence](http://opensource.org/licenses/MIT).`;
+This software is published by the Financial Times under the [MIT license](http://opensource.org/licenses/MIT).`;
```
JakeChampion:
```suggestion
This software is published by the Financial Times under the [MIT licence](http://opensource.org/licenses/MIT).`;
```

templates/component-boilerplate/read-me.js
sbinlondon: 🇺🇸 🙅 
🇬🇧 💃

templates/component-boilerplate/read-me.js
```diff
@@ -17,7 +17,7 @@ _A table of contents to help people find things_
 - [Contributing](#contributing)
 - [Migration](#migration)
 - [Contact](#contact)
-- [Licence](#licence)
+- [License](#license)
```
JakeChampion:
```suggestion
- [Licence](#licence)
```

lib/tasks/boilerplate.js
sbinlondon: @rowanmanning should we look at this tomorrow?

sbinlondon: Added! 😊

rowanmanning:
I think we're gonna leave silent mode documentation in while we have inconsistency. It's easy for someone to remove this if their component doesn't support silent mode, and more difficult if a new component _does_ support silent mode and we don't include this – they'd have to copy/paste from somewhere

gvonkoss:
I agree with the silent mode being in the docs, but init is for new components, and we're trying to move everything over to the primary mixin right now, so this would apply to anything that's up and coming in the pipeline, no?

PR #593, opened 2019-05-22
Test dart and node sass compilation.

notlee:
As Dart Sass is the primary implementation of Sass, we look to
use Dart Sass in a future major of obt or [oat](https://github.com/financial-times/origami-authoring-tools).

This commit adds Dart Sass compilation tests so we can be sure
our components compile with both Node and Dart Sass.

Screenshot 2019-05-22 at 13 56 06

Sass warnings are not logged as part of these tests as I removed the `@warning` custom function from compilation tests, as it caused dart-sass to error. Seems it was meant for custom Sass functions `error('my error');` not annotations `@error 'my error'`. I don't think this is a bad thing -- we should work to minimise component warnings.


notlee:
Somethings not quite right. Please hold.

PR #596, opened 2019-07-03
Prevent node Sass bundling with bower Sass.

notlee:
The test "should not compile sass which uses files in node_modules
directory" fails consistently when I run it locally. It has also
failed intermittently in CircleCI, see:
https://github.com/Financial-Times/origami-build-tools/runs/154454659

It appears the failure is correct as `getSassIncludePaths` includes
both "node_modules" and "bower_components", which means both bower
Sass and npm Sass could be bundled together.

This updates `getSassIncludePaths` to conditionally include
either bower or npm Sass according to the `--ignore-bower` flag.

However... if this is all true, it's not clear to me based on git history how the 
test has been passing previously 🤔


package-lock.json
```diff
@@ -11105,7 +11105,7 @@
       "dependencies": {
         "source-map": {
           "version": "0.4.4",
-          "resolved": "https://registry.npmjs.org/source-map/-/source-map-0.4.4.tgz",
+          "resolved": "http://registry.npmjs.org/source-map/-/source-map-0.4.4.tgz",
```
chee: what's happening here

notlee: 🤷‍♂ npm install with `npm@6.9.0` and `node@10.16.0` I think

PR #601, opened 2019-08-12
Remove `Array.from` from the boilerplate JS.

notlee:
We support IE11, which does not support `Array.from` without
a polyfill.


JakeChampion:
I think it might be better to use Array.from and have it in the required browserFeatures array.
Either Array.from or use `for..of` loop and add `NodeList.prototype.@@iterator` to the required browserFeatures array. I think that is a better approach because it embraces modern built-in features which the majority of our users have.

notlee:
Fair and valid (done)

PR #607, opened 2019-08-28
Use polyfill.io v3 in page template

chee:


chee:
I suppose this would be a breaking change

notlee:
Nice 👌 patch. Looks like demos are ok to use the default behavior of v3.
https://github.com/Financial-Times/polyfill-service/blob/master/MIGRATION.md#migrating-from-v2-to-v3

PR #610, opened 2019-09-25
Do not add webpack Bower config if ignoreBower is true

JakeChampion:


GlynnPhillips:
Could add a test as well https://github.com/Financial-Times/origami-build-tools/blob/master/test/unit/config/karma-config.test.js

PR #612, opened 2019-10-03
Make Sass errors more useful.

notlee:
Before
Screenshot 2019-10-03 at 12 23 03

After
Screenshot 2019-10-03 at 12 22 43


JakeChampion:
Nice!

PR #669, opened 2020-03-13
Only build Sass once when shared by multiple demos.

notlee:
Builds are asynchronous, and would previously all start before
the flag which prevents multiple of the same build was set.

o-buttons, which has a lot of demos, builds around ~4.5s instead
of ~36.65s. Making it about ~7.85 times faster! :ta-da:

### DIFF
--- a/lib/tasks/demo-build.js
+++ b/lib/tasks/demo-build.js
@@ -15,57 +15,53 @@ const fileExists = file => denodeify(fs.open)(file, 'r').then(() => true).catch(
 const readFile = denodeify(fs.readFile);
 const outputFile = denodeify(fs.outputFile);
 
-const builtFiles = {};
-
 function buildSass(buildConfig) {
 	const src = path.join(buildConfig.cwd, '/' + buildConfig.demo.sass);
 	const dest = path.join(buildConfig.cwd, '/demos/local/');
-	if (!builtFiles.css.includes(src)) {
-		return fileExists(src)
-			.then(exists => {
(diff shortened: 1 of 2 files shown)


lib/tasks/demo-build.js
```diff
@@ -415,35 +405,66 @@ module.exports = function (cfg) {
...
+				typeof config.watching === 'undefined' ||
+				config.watching === 'sass'
+			);
+			// Check if obt is watching for JavaScript changes.
+			const watchingJs = config && (
+				typeof config.watching === 'undefined' ||
+				config.watching === 'js'
+			);
+			// Create an array of configuration for each demo asset to build.
+			const htmlBuildsConfig = []
```
JakeChampion:
```suggestion
			const htmlBuildsConfig = [];
```

lib/tasks/demo-build.js
```diff
@@ -415,35 +405,66 @@ module.exports = function (cfg) {
...
+			const sassBuildsConfig = []
```
JakeChampion:
```suggestion
			const sassBuildsConfig = [];
```

lib/tasks/demo-build.js
```diff
@@ -415,35 +405,66 @@ module.exports = function (cfg) {
...
+			const jsBuildsConfig = []
```
JakeChampion:
```suggestion
			const jsBuildsConfig = [];
```

PR #676, opened 2020-03-20
move to node 12 and update dependencies

JakeChampion:
- Replaced Denodeify with util.promisify
Updated:
- ft/sass
- autoprefixer
- chalk
- chokidar
- colors
- eslint plugins but not eslint as it was harder to migrate
- fs-extra
- mocha
- node-sass
- pa11y
- portfinder
- rimraf
- makefile
- node-version


notlee:
Nice! Shall we remove `babel-plugin-add-module-exports` as part of this? Since it's a major release

JakeChampion:
> Nice! Shall we remove `babel-plugin-add-module-exports` as part of this? Since it's a major release

Yes, can we do it in another PR? I've made a v10 branch which we can PR all the breaking changes into and then we can make a PR from v10 to master and label that `release:major`

PR #687, opened 2020-03-20
migrate from serve to latest serve-handler

JakeChampion:
serve no longer has a API, it is only a terminal command now. I've moved us to the API package which is called serve-handler.

I've kept the same look and feel of serve but not added the copy link to clipboard feature. I can add that if we want it.


notlee:
>I've kept the same look and feel of serve but not added the copy link to clipboard feature. I can add that if we want it. 

I didn't know that was a thing until very recently -- because it got in my way where I wasn't expecting it in my clipboard! So happy to drop it personally

PR #690, opened 2020-03-20
Remove snyk ignore

notlee:


notlee:
let's update this after -- we at least need to keep the patch, unless a karma update solves that

PR #708, opened 2020-03-23
update parse-gitignore to latest version

JakeChampion:
parse-gitignore v1 breaking change is that it no longer attempts to turn ignores into globs.

We never used that behavior though :-)


notlee:
Our current components don't have `eslint-ignore.js`. Why did updating `parse-gitignore` require it as part of the boilerplate now?

PR #717, opened 2020-03-25
Replace `request-promise-native` with `node-fetch`.

notlee:
It is deprecated:
https://github.com/request/request-promise-native#deprecated


notlee:
> 👍
> I think we should take a look at replacing node-fecth with axios as it offers a better API and nice plugins which can automate retrying failed requests and listening to rate-limits

Noted! It does look like a better api. Moving to axios would need the current tests to be more significantly rewrote, so I'll leave this for the time being.

notlee:
Ugh but maybe I will -- mocking 'node-fetch' is proving a pain

PR #719, opened 2020-03-26
Replace babel with swc-loader, for improved performance.

notlee:


chee:
exciting

notlee:
Unscientific average of three `obt demo` builds on `o-forms`: 2.55s, down from 5.00s 🎉

chee:
make sure it can build `o-table`. any time i'm doing big changes, `o-table` always turns out to be the boss level

notlee:
I'm having problems when including node modules to support older browser versions. 

```
[./test/oTable.test.js] 3.52 KiB {test/oTable.test} [built]
    + 126 hidden modules
ERROR in ../origami-build-tools/node_modules/sinon/lib/sinon/proxy-call.js 110:20
Module parse failed: The keyword 'yield' is reserved (110:20)
File was processed with these loaders:
 * ../origami-build-tools/node_modules/swc-loader/src/index.js
You may need an additional loader to handle the result of these loaders.
|         throw this.args[pos];
|     },
>     yield: function yield() {
|         return this.yieldOn.apply(this, concat([
|             null
 @ ../origami-build-tools/node_modules/sinon/lib/sinon/mock-expectation.js 4:24-47
 @ ../origami-build-tools/node_modules/sinon/lib/sinon.js
 @ ./test/BasicTable.test.js
```

It looks like not a swc issue? A webpack issue?
````
➜  Desktop swc test.js -C jsc.target=es3
var a = {
    yield: function() {
        console.log('yield here');
    }
};
console.log(a.yield());
➜  Desktop swc test.js -C jsc.target=es2016
const a = {
    yield: ()=>{
        console.log('yield here');
    }
};
console.log(a.yield());
```

I couldn't figure it out in a timely fashion. Instead let's wait for oax to replace webpack.

PR #720, opened 2020-03-30
Move from Webpack to oax

JakeChampion:


lib/karma-oax.js
```diff
@@ -0,0 +1,50 @@
...
+		if (!config.ignoreBower) {
+			oaxArguments.push(`--for-bower`);
+		}
+
+		const oaxCommand = oaxPath + ' ' + oaxArguments.join(' ');
+		try {
+			log.info('Generating bundle for ./%s', location);
+			const {stdout, stderr} = await execa.command(oaxCommand, {
+				cwd: config.basePath
+			});
```
chee: maybe we should add a runner to oax, so that the user doesn't have to build up the command themselves

PR #721, opened 2020-04-01
use karma-proclaim to make proclaim a global variable used in tests

JakeChampion:
instead of abusing webpack to make proclaim available via import/require.

https://github.com/Financial-Times/origami-build-tools/blob/cf219aa5029754feb0f98c63db87f2af4a439786/config/webpack.config.dev.js#L6-L9


JakeChampion:
This is meant for the current major, which is why it is merging with master

chee:
exciting !

chee:
> tada This PR is included in version v9.0.5 tada

wait what

PR #725, opened 2020-04-02
Remove production option as this tool is not used for production builds

JakeChampion:


notlee:
Wasn't the plan to improve component authoring tools within obt but keep it for building projects too untill we eventually break out oat? We use obt to build the reigstry and during the intro to origami workshop -- unsure how useful a revamped obt would be for our users building projects but it comes up from time to time

JakeChampion:
What commands are used by products? Is it only `obt build`? If so, that can be replaced by `oax` and `swc` for JS files and `sass` for the Sass files.

notlee:
Just build, with the watch flag also

chee:
i'm not sure about removing build in v10.

maybe we should explore alternatives for the workshop first, and for the apps that still use it

JakeChampion:
Here's an alternative for an Origami application --> https://github.com/Financial-Times/origami-registry-ui/pull/271

notlee:
Agree with @chee! That example looks good Jake. If we wrap that up into a friendly little cli, I'd be happy to remove build in v10 :)

notlee:
Some tests need updating to remove sourcemap expectations

PR #727, opened 2020-04-03
migrate from sasslint to stylelint

JakeChampion:


notlee:
Nice work! :D We have sass-lint comments in our components though to disable/enable rules where appropriate, e.g. `// sass-lint:disable no-vendor-prefixes`. Does stylelint look for those or is `obt verify` going to start failing on our components?

notlee:
Yeah we'd need to update all out components to use [`stylelint-disable`](https://stylelint.io/user-guide/ignore-code) over `sass-lint:disable`. I think that's ok if we release a major of `origami-ci-tools` which includes obt v10. We can update all our components Sass comments when we upgrade `origami-ci-tools`. If we need to backport a feature to a previous major of a component obtv9 will continue to build and lint them like usual. Might cause some confusion though
if we're using obtv10 locally and obtv9 in CI for old components, as linting error will vary

What do you think @JakeChampion?

JakeChampion:
Sounds good with me

test/unit/tasks/verify-sass.test.js
```diff
@@ -389,10 +396,20 @@ describe('verify-sass', function () {
...
+      					'no-empty-rulesets/valid.scss:2:6',
+      					'no-empty-rulesets/valid.scss:9:7',
+      					'no-empty-rulesets/valid.scss:12:8'
+					]
+				},
 			];
 			return verify().task().then(() => {
 				throw new Error('No linting errors were thrown.');
 			}, error => {
+					console.log(error.message);
```
JakeChampion, line 412:
```suggestion
				console.log(error.message);
```

test/unit/tasks/verify-sass.test.js
```diff
@@ -12,7 +12,7 @@ const obtPath = process.cwd();
 const oTestPath = 'test/unit/fixtures/verify';
 const verifyTestPath = path.resolve(obtPath, oTestPath);
 
-describe('verify-sass', function () {
+describe.only('verify-sass', function () {
```
JakeChampion, line 15:
```suggestion
describe('verify-sass', function () {
```

test/unit/tasks/verify-sass.test.js
notlee: this is to help me. not done yet

test/unit/tasks/verify-sass.test.js
notlee: this is to help me. not done yet

notlee:
@Financial-Times/origami-core teamwork! This is ready for review now. 

**I recommend reviewing this with whitespace ignored**. 

Most of our `sass-lint` rules have been covered by `stylelint` and the plugins `stylelint-scss` and `stylelint-order`. There are some rules which seem to be missing but I don't think it's worth writing a rule for ourselves (let me know if you think different though). Exceptions should be clear from the test files. An example is linting for spaces around mixin and function parameters. Also the `stylelint` "no vendor prefix" rule does not apply to all vendor prefixes but only to those standard properties/values Autoprefixer would handle.

No rules log a warning anymore, as obt always exited with an error regardless.

config/.stylelintrc.json
```diff
@@ -0,0 +1,91 @@
+{
+   "plugins": [
+      "stylelint-scss",
+      "stylelint-order"
+   ],
+   "rules": {
+      "scss/at-import-no-partial-leading-underscore": true,
+      "scss/at-import-partial-extension": "never",
```
JakeChampion, line 8: This rule causes this line to fail in the template -- https://github.com/Financial-Times/origami-build-tools/blob/ee546af0e0dff8fd5e7dfcb900fa23186663ad1e/templates/component-boilerplate/test-sass-main.js#L9

JakeChampion:
```
Error - Unexpected extension ".test" in @import (scss/at-import-partial-extension)
```

config/.stylelintrc.json
```diff
@@ -0,0 +1,91 @@
...
+      },
+      "property-no-vendor-prefix": true,
+      "value-no-vendor-prefix": true,
+      "selector-no-vendor-prefix": true,
+      "media-feature-name-no-vendor-prefix": true,
+      "at-rule-no-vendor-prefix": true,
+      "function-comma-space-after": "always",
+      "function-parentheses-space-inside": "never",
+      "indentation": "tab",
+      "max-nesting-depth": 3,
```
JakeChampion, line 77: This rule causes the sass test files to fail linting:
https://github.com/Financial-Times/origami-build-tools/blob/ee546af0e0dff8fd5e7dfcb900fa23186663ad1e/templates/component-boilerplate/test-sass-mixins.js#L8
```
 Error - Expected nesting depth to be no more than 3 (max-nesting-depth)
```

notlee:
@JakeChampion this is not ideal
https://github.com/Financial-Times/origami-build-tools/pull/727/commits/ab7b0e392596f65dffb12501e126ce51b4cbab3b

Stylelint throws an incorrect `indent` error for Sass when a semicolon follows a mixin `@include`. Tried to avoid confusion by not allowing extra semicolons.

notlee:
Slightly messed up a rebase there, but let's roll with it https://github.com/Financial-Times/origami-build-tools/pull/727/commits/9d9008dcaee632bf67d5b4caca11cf6de981e114

PR #732, opened 2020-04-16
Always output Sass debug messages with the verbose flag.

notlee:
We were using our `log.debug` method to output Sass debug and
warning messages, which is only logged when the enviroment variable
`OBT_LOG_LEVEL` is set to `debug`. We are not debugging the build
tools themselves and so should use the `log.secondary` method
instead.


JakeChampion:
:tada: This PR is included in version v9.0.10 :tada:

PR #740, opened 2020-04-22
Move v10 from CircleCI to GH Actions

JakeChampion:


notlee:
>./test/oComponentBoilerplate.test.js:34:1 Error - Trailing spaces not allowed. (no-trailing-spaces)

The linter in the `obt install && demo && build && verify && test` integration test is catching boilerplate issues for the first time?

JakeChampion:
Odd that obt tests started to fail when I moved it to GH Actions
I fixed those issues now though
We have an issue in ft/sass where the generated sourcemaps are not reproducible across different operating systems

Issue #735, opened 2020-04-17
proposal: update obt v10 interface


notlee:
We decided to make obt v10 a specific tool for authoring components in development.
Where we use obt for non-components we are [replacing obt with other tools](https://github.com/Financial-Times/origami/issues/47).

Jake has opened [a PR to remove the production flag](https://github.com/Financial-Times/origami-build-tools/pull/725). That's great, and I would
like to propose that we take the opportunity to introduce other interface changes.

When working on a component I use the following command. Usually I end up
sharing this with other engineers when they start working on a component for
for the first time (there are a lot of options to get your head around):
`obt demo --run-server --watch --verbose --brand=master`

I tend also to use:
`obt test`
`obt verify`

And if there's a tricky JS test failure that needs some debugging:
`obt test --debug`

We should review the interface more thoroughly, I have a few more ideas,
with more input from engineers outside the team. But I'm keen not to 
endlessly increase the scope of v10. For now I propose quick wins that mostly 
strip away options we do not need:

- Remove any flag which isn't compatible with the Origami spec (i.e. our sass/js entry points) or any flag that isn't necessary when working on a component (build destination flags). 
- Default to and remove`--verbose` from all commands. Do not replace with `--quite`. In development or ci builds we always want to see warning and debug logs. 
- Remove ` --run-server --watch` from the `demo` command. Keep the demo command for building demos in ci as usual.
- Add a `dev` command with one `brand` flag which behaves the same as `obt demo --run-server --watch --verbose --brand=master`
- Default to and remove the `--suppress-error` flag, which we use with `--demo-filter` [in ci](https://github.com/Financial-Times/origami-ci-tools/blob/f16d33b3c266356484f52c70d7f50c522ff8d5b2/commands/branch.js#L29) to prevent errors when the filtered demo isn't found.

```diff
	Usage
		$ obt <command> [<options>]

	Commands
+		develop, dev  Build demos locally every time a file changes and run a server to view them.
		build, b      Build CSS and JS in current directory
		demo, d       Build demos into the demos directory
		init          Initialise a new component with a boilerplate folder structure
		install, i    Install npm and bower dependencies required to build modules
		test, t       Run Origami specification tests and component specific tests
		verify, v     Check folder and code structure follows Origami specification

	Options
		-h, --help                 Print out this message
-		--watch                    Re-run every time a file changes
-		--run-server               Build demos locally and runs a server
-		--js=<path>                Main JavaScript file (default: ./src/main.js)
[...]

PR #741, opened 2020-04-23
Update uglifyJs plugin for webpack

alexflorisca:
This solves an issue where the uglifyJs webpack plugin would crash when it saw ES2015 code, as detailed here: https://github.com/webpack/webpack/issues/5858

To reproduce this error, try to run `obt build --production` on a js file that has the following code: 
```js
const b = 'hey'
const a = { b }
```
I came about this error trying to package the npm build of o-viewport here: https://app.circleci.com/pipelines/github/Financial-Times/o-ads/19/workflows/09cf71cd-38e9-40a8-9b16-dccb0cfd607f/jobs/2812

New options for UglifyJs are here: https://github.com/mishoo/UglifyJS2#minify-options

### DIFF
--- a/config/webpack.config.prod.js
+++ b/config/webpack.config.prod.js
@@ -1,5 +1,6 @@
 const webpack = require('webpack');
 const webpackMerge = require('webpack-merge');
+const UglifyJsPlugin = require('uglifyjs-webpack-plugin');
 
 const prodConfig = {
 	plugins: [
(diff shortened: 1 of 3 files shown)


notlee:
Thanks Alex! I'd be happy to merge this except we're planning on removing the `--production` flag soon. Details here: https://github.com/Financial-Times/origami-build-tools/issues/735

Is there a particular reason you publish built files?

alexflorisca:
Hi @notlee we don't publish the built files, but we do run a bundle size checker as part of our build, which is where we use the `--production` flag.

alexflorisca:
But we get why you are making the change to remove the flag, and we are putting o-ads into maintenance mode, with a view to build other non-origami components to replace it, so we will remove the --production flag and disable the bundle check once the final big change has been made.

PR #743, opened 2020-04-27
Lint tests

notlee:


test/unit/fixtures/verify/src/scss/class-name-format/valid.scss
```diff
@@ -0,0 +1,13 @@
...
+  content: '';
+}
+
+.snake_case {
+  content: '';
+}
+
+// we allow leading with an underscore,
+// but maybe shouldn't for css class names
+._with-leading-underscore {
```
JakeChampion, line 11: 🤔 I wonder if we supported this by accident

notlee: probably, there's quite a few things like this we might want to tidy up

notlee:
```
          should not error if there are no Sass files:
     Error: ENOENT: no such file or directory, chdir '/home/runner/work/origami-build-tools/origami-build-tools/test/unit/fixtures/verify' -> './src/js'
```
getting this in ci but not locally. unsure why

PR #750, opened 2020-04-28
Remove --suppress-errors flag and don't throw when no demos are found

JakeChampion:


lib/tasks/demo-build.js
```diff
@@ -353,17 +353,6 @@ module.exports = function (cfg) {
...
-				if (config && demoFilters) {
-					noDemosError += ' for ' + String(demoFilters);
-				}
-				noDemosError += '.';
-
-				const e = new Error(noDemosError);
-				e.stack = '';
-				throw e;
-			}
 			// Check if obt is watching for Sass changes.
```
chee, line 356: does it just silently do nothing if the person has asked to run demos and they don't exist?

JakeChampion: Only if the --demo-filter flag is also passed
```
❯ ../../../origami-build-tools/lib/origami-build-tools-cli.js demo
 ✖ Compiling Demos   
    → Sass file not found: /Users/jake.champion/Code/repo-data-cli/components/o-table… 
❯ ../../../origami-build-tools/lib/origami-build-tools-cli.js demo --demo-filter=jake
✔ Compiling Demo
```

README.md
```diff
@@ -121,6 +120,7 @@ obt init o-my-new-component
 
 - NodeJS v10 is no longer supported. Use NodeJS v12 or above.
 - A default CommonJs export now maps to `module.exports.default`, the default [Babel](https://babeljs.io/) behaviour. If using `require` to include a default CommonJs export add a `.default` property to the `require` call. Alternatively update your project to use [ECMAScript Module syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules).
+- The `--suppress-errors` flag has been removed. OBT no longer throws an error if their are no demos to be built.
```
JakeChampion, line 123:
```suggestion
- The `--suppress-errors` flag has been removed. OBT no longer throws an error if their are no demos to be built if passed the `--demo-filter` flag.
```

lib/tasks/demo-build.js
chee: hm, i wonder if it would be nice to print something like "skipped: no demos matched the given filter"?

JakeChampion: Yes, I like that idea a lot!

JakeChampion: Done!

PR #751, opened 2020-04-28
Remove flags we no longer want to support

JakeChampion:
This is part of the work for https://github.com/Financial-Times/origami-build-tools/issues/735

- Removed the `--js` flag.
- Removed the `--sass` flag.
- Removed the `--build-js` flag.
- Removed the `--build-css` flag.
- Removed the `--build-folder` flag.
- Removed the `--standalone` flag.
- Removed the `--verbose` flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,6 @@ If you have any issues with OBT, please check out [troubleshooting guide](https:
 		-h, --help                 Print out this message
 		--watch                    Re-run every time a file changes
 		--run-server               Build demos locally and runs a server
-		--js=<path>                Main JavaScript file (default: ./src/main.js)
 		--sass=<path>              Main Sass file (default: ./src/main.scss)
 		--build-js=<file>          Compiled JavaScript file (default: main.js)
 		--build-css=<file>         Compiled CSS file (default: main.css)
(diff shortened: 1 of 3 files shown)


notlee:
Nice! I think it would be a good opportunity to simplify obt and remove code related to supporting those flags, as well as the flags themselves?

JakeChampion:
@notlee Which commands are you thinking about? I think all the commands still take flags. The `--ignore-bower` flag is used in the `install`, `test` and `build` commands. The `--run-server` and `--watch` flags are used by all the commands as well

chee:
could remove this ternary, and the Promise.resolve wrapper https://github.com/Financial-Times/origami-build-tools/blob/044b02c9e230c6adb5fd0b1e10b5f140c325263c/lib/tasks/build-sass.js#L59
and these config.buildCss and config.buildFolder https://github.com/Financial-Times/origami-build-tools/blob/044b02c9e230c6adb5fd0b1e10b5f140c325263c/lib/tasks/build-sass.js#L63-L64

chee:
and this if statement https://github.com/Financial-Times/origami-build-tools/blob/aaff645da2aaebc90cbf355d95d9929e7bb986ba/lib/tasks/build-js.js#L51-L53

JakeChampion:
@chee Those are used by the `obt demo` command, we can't remove them unfortunately.

JakeChampion:
> and this if statement
> 
> https://github.com/Financial-Times/origami-build-tools/blob/aaff645da2aaebc90cbf355d95d9929e7bb986ba/lib/tasks/build-js.js#L51-L53

We are not using webpack in v10, I think this is from the previous major of obt

chee:
oh that's a pity

PR #755, opened 2020-04-30
make obt init call npm init origami-component

JakeChampion:


notlee:
Nice! We can delete all the stuff in ` lib/tasks/boilerplate`

notlee:
And all of these 😄 https://github.com/Financial-Times/origami-build-tools/pull/755/commits/bbbd520a7895d06ec31f7dfd57f45f79864e0f00#diff-f4ff27affcdccb91e156216ec206fa9cL11-L34

notlee:
There's an integration test that runs `obt  init` then checks building, demos, and linting doesn't error. That'll fail as the cli is interactive now. Maybe we can delete that test in favour of a unit test which checks create-origami-component is called? Since create-origami-component is responsible for creating a new component now

JakeChampion:
I'm fixing all these things :-)

chee:
ha! fantastic

chee:
yay! this is fantastic

PR #760, opened 2020-04-30
Fix JavaScript build for `dev` and `demo` commands.

notlee:


lib/tasks/build-js.js
```diff
@@ -39,7 +39,9 @@ module.exports = async function (cfg) {
 
 		const swcArguments = ['--filename', '-', '--no-swcrc', '-C', 'minify', '--source-maps', 'inline'];
 
-		const swc = run('swc', swcArguments, dontOutputToTerminal);
+		const swc = run('swc', swcArguments, Object.assign({
+			localDir: path.resolve(__dirname, '../../node_modules/.bin')
```
JakeChampion, line 43: This shouldn't be needed :/ we use execa and execa adds the node_modules/.bin to the path I thought?

notlee: without digging I can only add ¯\_(ツ)_/¯ ...works now, didn't before

notlee: I'm going to write an integration test for the demo command

JakeChampion: I don't know if that will work for globally installed obt though, is there a way to test?

notlee: we're using the same pattern already for bower
		localDir: path.resolve(__dirname, '../../node_modules/.bin')

notlee: https://github.com/sindresorhus/execa/issues/169

notlee: tests over here https://github.com/Financial-Times/origami-build-tools/pulls

PR #766, opened 2020-05-04
Always show Sass debug and warning notices.

notlee:
We removed the `--verbose` flag but did not set `verbose: true`
by default. In other words, the `verbose` option still existed
but had no cli flag to set it to true.

Remove the verbose option all together and update our unit tests
to always expect debug output from Sass builds


origamiserviceuser:
:tada: This PR is included in version v10.0.2 :tada:

PR #767, opened 2020-05-04
Move `stylelint-order` to `dependencies` from `devDependencies`

notlee:
**Move `stylelint-order` to `dependencies` from `devDependencies`**
`obt verify` errors when installed globally, as `stylelint-order` cannot be found. I assume this is why.

**Remove `is-es5-syntax` as a dev dependency, as it is not used.**


JakeChampion:
we run run our integration tests against the `npm pack`ed version of this project, that would catch this type of bug

notlee:
hmm, tests are timing out, that's annoying

origamiserviceuser:
:tada: This PR is included in version v10.0.3 :tada:

PR #770, opened 2020-05-05
Use component stylelint rules if defined.

notlee:
- Check the component directory for stylelint rules and use
those if found.
- Extend the `stylelint-config-origami-component` rules by
default.

On linting Sass, our specification says:
>Sass must validate using the Origami Sass Lint rules, though
>exceptions may be enabled temporarily within a component using
>Sass Lint comments.
>
>Component indent type (tabs or spaces) is not standardised:
>developers must respect whatever indent type is already in use
>when editing existing components.

In practice it was difficult to match the existing indentation of
a component and we often ended up with a mix. Origami Build Tools
v10 therefore lints for tab indentation. Enabling components to
extend `stylelint-config-origami-component` and individually
customise Stylelint config will allow engineers who build Origami
components to choose spaces if they feel strongly. It also helps
text editor integration.


origamiserviceuser:
:tada: This PR is included in version v10.1.0 :tada:

PR #776, opened 2020-05-06
Update Sass linting to allow newlines in declarations and functions.

notlee:
This is useful for formatting functions with large arguments in a
readable way:
```scss
$some-very-very-very-large-conditional: true;
$foreground-color: if(
	$some-very-very-very-large-conditional == true,
	"then each argument should go on a newline for readability",
	"otherwise each argument could be on the same line"
);
```

And for formatting declarations such as grid areas in a readable
way:
```scss
.foo-grid {
	grid-template-areas:
		"header header header"
		". sidebar ."
		". main ."
		"footer footer footer";
}
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3153,7 +3153,7 @@
     },
     "event-stream": {
       "version": "3.3.4",
-      "resolved": "https://registry.npmjs.org/event-stream/-/event-stream-3.3.4.tgz",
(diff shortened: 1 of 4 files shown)


origamiserviceuser:
:tada: This PR is included in version v10.1.5 :tada:

PR #777, opened 2020-05-06
Fix the `--brand` flag when used with the `dev` command.

notlee:


lib/origami-build-tools-cli.js
```diff
@@ -179,10 +182,10 @@ if (task && (argument === 'develop' || argument === 'dev')) {
 				childProcess.kill();
 			}
 
-			childProcess = runWatchedTask(['demo']);
+			childProcess = runWatchedTask(devArguments);
```
JakeChampion, line 185: is it possible to have a test for this?

notlee: I opened an issue. We should add tests for the whole command https://github.com/Financial-Times/origami-build-tools/issues/778

origamiserviceuser:
:tada: This PR is included in version v10.1.6 :tada:

PR #808, opened 2020-06-12
Make obt output GitHub Annotations for errors and warnings

JakeChampion:
By logging out in the format:
```js
console.log(`::error file=${file},line=${line},col=${column}::${message}`);
```
GitHub will add an annotation to the `file` at the `line` and `col` location with the contents of `message`.

We can also url-encode any new-line characters in the `message` and the new-line characters will appear in the annotation as well.

Approaching GitHub annotations this way instead of using [GitHub Problem Matchers](https://github.com/actions/toolkit/blob/master/docs/problem-matchers.md) is a better solution for Origami because it means we have a single location where the code for the annotations lives instead of being spread across different repos and files. It also is far simpler to understand than a JSON file which contains Regular Expressions and also allows us to have different output in CI and in the annotation itself :-)


lib/tasks/pa11y.js
```diff
@@ -28,21 +28,20 @@ const pa11yTest = async function (config) {
...
-				'\n' +
-				'   ├── ' + result.selector.replace(/\s+/g, ' ') +
-				'\n' +
-				'   └── ' + result.context.replace(/\s+/g, ' ')).grey
-		);
-
 	if (results.issues.some(result => result.type === 'error')) {
-		throw new Error(errors.join('\n') + '\nFailed Pa11y tests');
+		// Process and return the results
+		require('colors');
```
notlee, line 33: oh what does this `require('colors');` do?

JakeChampion: It adds lots of methods to the String prototype 🙈 Let's fix this to use `colors/safe` instead

test/unit/plugins/listr-karma.test.js
```diff
@@ -98,7 +81,7 @@ describe('SpecReporter', function () {
 							failedSpec = {
 								suite: ['suite name'],
 								description: 'description of test',
-								log: ['log', 'test/add.test.js']
+								log: ['2+2 doest not equal carrot', 'test/add.test.js:1:12']
```
notlee, line 84: `2+2 doest not equal carrot` 😂

JakeChampion: Hey now, I'm almost certain that it does not equal carrot :P

origamiserviceuser:
:tada: This PR is included in version v10.5.0 :tada:

PR #810, opened 2020-06-15
Only add karma annotations when we know which file to add them to

JakeChampion:
When a test fails due to a timeout, we do not get given the file information for where the test originated from. Which means we can not add a good annotation to the pull-request


origamiserviceuser:
:tada: This PR is included in version v10.5.3 :tada:

PR #814, opened 2020-06-16
Add core/enhanced experience to Origami demos.

notlee:
We document setting up a core/enhanced experience here:
https://origami.ft.com/docs/components/compatibility/#core--enhanced-experiences

The instructions include adding this CSS snippet:
```css
.core .o--if-js,
.enhanced .o--if-no-js {
        display: none !important;
}
```

I  expect we should be able to use the `o--if-js` and `o--if-no-js`
class across Origami components. Like we once did in the deprecated
`o-section-head` component:
https://github.com/Financial-Times/o-section-head/blob/e6a3d25337b2e617a84a68d7e7edaceb99244442/main.scss#L52

However we do not set up a core/enhanced experience for our demos,
so we can't use these classes. Previously a `demo-js` class was
used instead but in 2016 `demo-js` css was removed. The core/enhanced
CSS was added to demos instead but without the JavaScript to add or
remove them. I assume this was a mistake (but I guess could have been
an attempt to allow components to demo a core experience even with JS
enabled by adding the `core` class -- maybe?)
https://github.com/Financial-Times/origami-build-tools/pull/339/files#diff-b174cbcda4fa263aeb31a92e25f4534bR9

Many if not all of our current components add there own
class if JS is enabled to achieve the same thing. e.g. `o-tabs`:
https://github.com/Financial-Times/o-tabs/blob/70f6bb62ed10d8fa31fc28f643f6c570e401ba43/src/js/Tabs.js#L6

This commit implements the core/enhanced mode for our demos so our
components may use the `o--if-js` and `o--if-no-js` classes instead.
https://origami.ft.com/docs/components/compatibility/#core--enhanced-experiences

The use of `staticSource` and the last `config.production` reference has
been removed as part of this work since obt v10 does not support a
production build or source options.

**_Aside: we will need to port this to obtv5 for the Build Service
also. My personal preference is actually for each component to
handle the core experience themselves so users have less to setup,
and components do not break if users do not include the
core/enhanced CSS. Perhaps worth creating a proposal to remove
the core/enhanced CSS requirement before merging this and doing the
Build Service work?_**


JakeChampion:
> My personal preference is actually for each component to
> handle the core experience themselves so users have less to setup,
> and components do not break if users do not include the
> core/enhanced CSS.

I agree with this and so does @chee :+1

JakeChampion:
@notlee Shall we close this PR based on the previous comment?

> > My personal preference is actually for each component to
> > handle the core experience themselves so users have less to setup,
> > and components do not break if users do not include the
> > core/enhanced CSS.
> 
> I agree with this and so does @chee 👍

notlee:
Okay I've updated this PR to remove the core/enhanced classes from obt demos. The demo cuts the mustard test remains in this PR, so when we test components in core experience browsers JS doesn't run and error.

notlee:
sorry @JakeChampion I had to [make changes](https://github.com/Financial-Times/origami-build-tools/pull/814/files/180de706d1671d5697b699fb4cc74a2615c5ab51) since your last approval

JakeChampion:
@notlee It's so odd, I saw those changes before my approval, I swear!

notlee:
A hanging build with no output. WHY GITHUB

origamiserviceuser:
:tada: This PR is included in version v10.5.8 :tada:

PR #861, opened 2020-07-27
add task for verifying readmes

chee:


lib/tasks/verify-readme.js
```diff
@@ -0,0 +1,81 @@
...
+
+	return {
+		title: "Verifying your component's README.md",
+		task: () => origamiJson(config),
+		skip: function() {
+			const origamiJsonPath = path.join(config.cwd, "/origami.json");
+
+			return fileExists(origamiJsonPath).then(exists => {
+				if (!exists) {
+					return `No README.md found. An Origami component requires a readme. Create ${path.join(
```
JakeChampion, line 73: I think this will make the task get skipped, what do you think about not having a skip function and instead making the task throw an error with this message instead? That way the verify task as a whole will fail if the readme file is missing

chee: ah. that makes sense. I copied this from the origami.json task, but i can see why that makes sense for origami.json and not for README.md

the task will already throw an error if the file isn't there, so i can remove this section and it will be just as good

chee:
:) it works https://github.com/Financial-Times/o-typography/pull/275/files#diff-04c6e90faac2675aa89e2176d2eec7d8

chee:
yay! it worked!

package.json
```diff
@@ -1,89 +1,96 @@
...
+		"@swc/cli": "^0.1.24",
+		"@swc/core": "^1.1.39",
+		"aliases": "^1.0.0",
+		"autoprefixer": "^9.7.6",
+		"bower": "^1.8.8",
+		"boxen": "^4.2.0",
+		"chalk": "^4.1.0",
+		"chokidar": "^3.4.0",
+		"colors": "^1.4.0",
+		"create-origami-component": "^0.2.13",
```
JakeChampion, line 25: It looks like this is a new dependency, I can't see it being used anywhere though. Is it being used?

chee: oh, weird. i guess we depend on it silently:

https://github.com/Financial-Times/origami-build-tools/blob/a1dd443e0fa0e9cf1fbd301eff91f0a6dd49c2d1/lib/tasks/init.js#L4-L6

we should probably lock that to a version? currently a change to `create-origami-component` can break working installs of `obt` and we wouldn't notice. maybe that's fine.

JakeChampion: oh! Nice work spotting that :-D Let's keep it defined in package.json, great idea 👍

origamiserviceuser:
:tada: This PR is included in version v10.6.0 :tada:

notlee:
nice work @chee :D

PR #862, opened 2020-07-30
Test on 12.x and 14.x

chee:
because [engines](https://github.com/Financial-Times/origami-build-tools/blob/master/package.json#L11) says we support 12+

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,11 +1,11 @@
-name: Test
+name: Test with node LTS and current
 on: [pull_request]
 jobs:
   build:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [10.x, 12.x]
+        node-version: [12.x, 14.x]
     steps:
     - uses: actions/checkout@v2
     - name: Use Node.js ${{ matrix.node-version }}


JakeChampion:
I think we need to keep 10 also as it is still an LTS release until June 2021 -- https://nodejs.org/en/about/releases/

chee:
> I think we need to keep 10 also as it is still an LTS release until June 2021 according to the [NodeJS releases page](https://nodejs.org/en/about/releases/)

oh, should we lower the `engines` to be 10+ as well then? https://github.com/Financial-Times/origami-build-tools/blob/master/package.json?rgh-link-date=2020-07-30T17%3A26%3A13Z#L11

JakeChampion:
> > I think we need to keep 10 also as it is still an LTS release until June 2021 according to the [NodeJS releases page](https://nodejs.org/en/about/releases/)
> 
> oh, should we lower the `engines` to be 10+ as well then? [`master`/package.json?rgh-link-date=2020-07-30T17%3A26%3A13Z#L11](https://github.com/Financial-Times/origami-build-tools/blob/master/package.json?rgh-link-date=2020-07-30T18%3A14%3A38Z#L11)

I think I've misunderstood what the change was meant to do, sorry, the name of the GitHub Workflow made me think we were wanting to test on the LTS versions of NodeJS as well as the current version -- [`a932e60` (#862)](https://github.com/Financial-Times/origami-build-tools/pull/862/commits/a932e60ab67f395397028444d4c211d9c19f35b5#diff-fe8421955fd596131bb6f1b78984b2fbR1)

Now I see we are testing on the versions which are specified in the package.json, which makes perfect sense 👍

PR #865, opened 2020-08-03
use our version of create-origami-component

### LINKED ISSUE #864: Defer to the depended on version of create-origami-component
https://github.com/Financial-Times/origami-build-tools/blob/a1dd443e0fa0e9cf1fbd301eff91f0a6dd49c2d1/lib/tasks/init.js#L4-L6

here we use `npm init origami-component`, which won't use the one that's in our deps. we want to change this to `npx create-origami-component` or `node node_mobuldes/.bin/create-origami-component`

chee:
closes #864

The previous version of `obt init` was using `npm init origami-component` so there was an invisible dependency.
now we depend on a version range explicitly so we can use that version.

### DIFF
--- a/lib/tasks/init.js
+++ b/lib/tasks/init.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const {run} = require('../helpers/command-line');
+
 module.exports = function () {
-	return run('npm', ['init', 'origami-component']);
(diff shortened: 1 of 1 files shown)


origamiserviceuser:
:tada: This PR is included in version v10.6.2 :tada:

PR #873, opened 2020-08-05
True removed support for node-sass in v6.

notlee:
Relates to:
https://github.com/Financial-Times/origami-build-tools/pull/856

### DIFF
--- a/lib/tasks/test-sass.js
+++ b/lib/tasks/test-sass.js
@@ -18,17 +18,18 @@ const trueTest = function(config, task) {
 		const componentBase = path.join(__dirname, '../../');
 		const sassFile = path.join(__dirname, 'index.test.scss');
 		const sassTrue = require('${require.resolve("sass-true")}');
+		const sass = require('${require.resolve("sass")}');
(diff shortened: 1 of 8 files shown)


lib/tasks/test-sass.js
```diff
@@ -18,17 +18,18 @@ const trueTest = function(config, task) {
 		const componentBase = path.join(__dirname, '../../');
 		const sassFile = path.join(__dirname, 'index.test.scss');
 		const sassTrue = require('${require.resolve("sass-true")}');
+		const sass = require('${require.resolve("sass")}');
```
JakeChampion, line 21: Could we use @financial-times/sass instead? That way we don't need to include two different versions of dart-sass

notlee: [Unfortunately not without changes to True](https://github.com/oddbird/true/blob/ba0c683c0d976a1f82e94c70fb7299509dcdcc01/lib/main.js#L53).

origamiserviceuser:
:tada: This PR is included in version v10.6.3 :tada:

PR #904, opened 2020-09-03
Pin styleint to a previous version.

notlee:
It appears a later version has introduced a regression.
In this `o-header` PR an ignore comment is not recognised.
```
transform: translate3d(-100%, 0, 0); // stylelint-disable-line declaration-block-no-duplicate-properties
```
https://github.com/Financial-Times/o-header/pull/395/

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
 		"serve-handler": "^6.1.2",
 		"sinon": "^9.0.2",
 		"snyk": "^1.319.0",
-		"stylelint": "^13.2.1",
+		"stylelint": "13.2.1",
 		"stylelint-config-origami-component": "^1.0.3",
 		"update-notifier": "^4.1.0",
 		"vfile": "^4.1.1",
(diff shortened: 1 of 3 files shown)


JakeChampion:
I guess we should look to have a test for this in obt to ensure any later updates to stylelint don't regress like this as well

package.json
```diff
@@ -14,7 +14,7 @@
 	"dependencies": {
 		"@financial-times/sass": "^1.31.0",
 		"@swc/cli": "^0.1.24",
-		"@swc/core": "^1.1.39",
+		"@swc/core": "^1.2.22",
```
JakeChampion, line 17: Let's only change stylelint in this pull-request and keep the other dependencies at the same version.

notlee:
> I guess we should look to have a test for this in obt to ensure any later updates to stylelint don't regress like this as well

already on it :)

package.json
notlee: ugh, choore

JakeChampion (approved):

origamiserviceuser:
:tada: This PR is included in version v10.7.3 :tada:

PR #905, opened 2020-09-04
add extension to relative imports to make the code work directly in a browser

JakeChampion:
relative imports require the full path including the extension because browsers use the import name to make a network request and they do not implicitly add a .js file extension if one is missing


chee (approved):

origamiserviceuser:
:tada: This PR is included in version v10.7.4 :tada:

Issue #906, opened 2020-09-07
feature: install demo dev dependencies instead of using the build service


notlee:
The feature request is lazy issue making, there might be other options, the problems to solve are:

- Currently demo dependencies listed in `origami.json` are included with a build service url along with a separate local build of the component. This means locally a demo may include a dependency which conflicts with a component dependency. When released the demo will then break unexpectedly and often goes unnoticed.

- The polyfills required by demo dependencies, as defined in the `origami.json` `browserFeatures` property, are not included in the demo html.

JakeChampion:
@Financial-Times/origami-core, would this be solved by the [proposal for convention over configuration for demos](https://github.com/Financial-Times/origami/issues/82)?

PR #907, opened 2020-09-07
Include the `CustomEvent` polyfill in all demos

notlee:
So demos with dev dependencies work in older browsers. This is because the
Origami Build Service includes `o-autoinit` which requires the `CustomElement`
polyfill.

This has come up because we stopped including the "default" set of polyfills
which included `CustomElement`.
https://github.com/Financial-Times/origami-build-tools/pull/623

We could solve this in a better way, issue here:
https://github.com/Financial-Times/origami-build-tools/issues/906


JakeChampion (approved):

origamiserviceuser:
:tada: This PR is included in version v10.8.0 :tada:

PR #912, opened 2020-09-23
compile the bundled code to es5 to work in IE11

JakeChampion:


config/karma.config.js
```diff
@@ -41,6 +41,7 @@ module.exports.getBaseKarmaConfig = function (opts = { ignoreBower: false }) {
 
 			// list of files / patterns to load in the browser
 			files: [
+				require.resolve('@babel/polyfill/dist/polyfill.js'),
```
notlee, line 44: We probably shouldn't include this in our tests as then our tests won't catch when we have forgotten to define a required polyfill in `origami.json` -- I think those polyfills are included?

JakeChampion: We need to in order to get async functions working which seem to require the regenerator runtime. I couldn't get it to work any other way.

JakeChampion:
Closing in favour of using swc instead of babel -- https://github.com/Financial-Times/origami-build-tools/pull/913

PR #913, opened 2020-09-23
When testing with Karma compile the  code to es5 to run in IE11

JakeChampion:
currently we are not compiling to ES5 and so the code will not execute in IE11


notlee (approved):

origamiserviceuser:
:tada: This PR is included in version v10.8.2 :tada:

PR #914, opened 2020-09-24
Add regenerator runtime to ensure async functions work when compiled to es5

JakeChampion:
Our compiler (swc) add `require('regenerator-runtime')` to it's output when compiling an async function. We need to add regenerator-runtime as a dependency to obt and include it in the bundle output for tests in order for them to run in older browsers such as IE11.


notlee (approved):

origamiserviceuser:
:tada: This PR is included in version v10.8.3 :tada:

PR #920, opened 2020-10-09
Pass through the ignoreBower flag to obt develop

chee:
So that it can work with a post-occ (or npm native :eyes:) origami component

### DIFF
--- a/lib/origami-build-tools-cli.js
+++ b/lib/origami-build-tools-cli.js
@@ -159,9 +159,17 @@ if (task && (argument === 'develop' || argument === 'dev')) {
 	startServer(port);
 	// build demos whenever changes are detected
 	let childProcess;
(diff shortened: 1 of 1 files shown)


notlee (approved):

notlee:
eventually we should check `bower.json` for the version and use that, so we don't have to remember to use the right flag :D

chee:
this actually won't work, i think we should try creating a demo bundle entirely via npm if `origamiVersion == 2.0.0`

JakeChampion:
> this (my PR) actually won't work (as you pointed out [here](https://github.com/Financial-Times/origami/pull/86/files#diff-624dbebdb0453398be5556ca97936012R54)) , i think we should try creating a demo bundle entirely via npm if `origamiVersion == 2.0.0`

Should we close this PR?

PR #923, opened 2020-10-29
do not call npm test from within obt test

JakeChampion:
by having `obt test` call `npm test` we are stopping our components from having `obt test` within their `npm test` script.

If we stop calling `npm test` from within `obt test` then we can can add `obt test` within each components `npm test` script, thereby making `npm install && npm test` work for our components.

I think I've worded this poorly, sorry everyone.

With the current obt, this would be an infinite loop:
```json
{
	"scripts": {
		"test": "obt test"
	}
}
```

If this pull-request is merged then the above configuration would work.

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -1,5 +1,9 @@
 # Migration Guides
 
+## Migrating from v10 to v11
+
+`obt test` no longer runs `npm test`, this is to allow `obt test` to be a command within the components `npm test` script.
+
 ## Migrating from v9 to v10
 The following `demo` command flags have been removed and replaced with the `develop` (`dev`) command:
 - Removed the `--watch` flag.
--- a/README.md
+++ b/README.md
@@ -108,7 +108,6 @@ Runs JavaScript and Sass tests.
 
 Checks Sass supports [silent and non-silent compilation modes](https://origami.ft.com/spec/v1/sass/#sass-silent-mode).
 If `pa11y.html` demo exists, confirms it is accessible using [Pa11y](http://pa11y.org/).
-If `package.json` contains a `test` script, confirms it exits with a 0 exit code.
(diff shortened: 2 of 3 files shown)


notlee (approved):
I think this is a breaking change, since we document that custom tests are allowed. Practically I can't think of a place it would be relied on outside [o-ads](https://github.com/Financial-Times/o-ads/blob/5a3ce35a3dfbc7e2b4a4c97aafcadc126e690a25/package.json#L9)

JakeChampion:
With the current obt, this would be an infinite loop:
```json
{
	"scripts": {
		"test": "obt test"
	}
}
```

If this pull-request is merged then the above configuration would work.

chee:
> I think this is a breaking change, since we document that custom tests are allowed. Practically I can't think of a place it would be relied on outside [o-ads](https://github.com/Financial-Times/o-ads/blob/5a3ce35a3dfbc7e2b4a4c97aafcadc126e690a25/package.json#L9)

true, but we don't actually document that we will run them for you with obt. custom tests are still allowed, you just have to type "npm test".

that being said, if we want to do this without changing it for people who are relying on obt running `npm test`, then we can instead check the environment variable `npm_lifecycle_event`. if `npm_lifecycle_event` is unset, (or is not `test`), then we run `npm test`.

https://docs.npmjs.com/cli/v7/using-npm/scripts#current-lifecycle-event

notlee:
Good point chee! And on closer inspection looks like [o-ads isn't using obt test](https://github.com/Financial-Times/o-ads/blob/5a3ce35a3dfbc7e2b4a4c97aafcadc126e690a25/.circleci/config.yml#L16) after all. I'm happy to release this as a patch, and revert if we need to. Or go with an `npm_lifecycle_event` check as Chee suggests (great suggestion)

JakeChampion:
I found one line of documentation for the feature -- https://github.com/Financial-Times/origami-build-tools/pull/923/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L111

We could roll this in with other types of breaking changes such as #924 and #925 and only have these features/feature-removals in a new version of obt which does not support spec v1 components, what do you both think?

PR #924, opened 2020-10-30
Remove bower support from obt entirely

JakeChampion:
@Financial-Times/origami-core what do you think about removing bower support completely from within obt?

The current version of obt will be used for origami spec v1 components and this new version of obt can be used for origami spec v2 components.


notlee:
Yes let's do this and have each component include `origami-build-tools` as a dev dependency to allow us to switch between v1 and v2 components. We should optimise obt to reduce the install time/size if each component is going to include its own copy

lib/tasks/demo-build.js
```diff
@@ -25,14 +25,14 @@ function buildDemoSass(buildConfig) {
...
+			return files.getPackageJson(buildConfig.cwd);
 		})
-		.then(bowerConfig => {
+		.then(pkgJson => {
 			let prefixSass = '';
 			// If module has o-assets as a dependency, set local demos to use local assets
 			// to use the latest ones in the repo
-			if (bowerConfig && bowerConfig.dependencies && bowerConfig.dependencies['o-assets']) {
-				const moduleName = bowerConfig.name;
+			if (pkgJson && pkgJson.dependencies && pkgJson.dependencies['o-assets']) {
```
notlee, line 34: I think we should check `peerDependencies` here? Or remove it it since [o-assets isn't used and will no longer work](https://github.com/Financial-Times/origami/pull/102)?

chee:
you know, maybe we should move `origami-ci-tools` into obt

lib/tasks/demo-build.js
JakeChampion: yes, let's remove this as we won't be bringing o-assets over to npm for the forseeable future.

notlee:
> you know, maybe we should move `origami-ci-tools` into obt

I'm not opposed to that. A publish command in obt would make sense.

The bundle size part of `origami-ci-tools` should probably be its own Github action.

chee:
> > you know, maybe we should move `origami-ci-tools` into obt
> 
> I'm not opposed to that. A publish command in obt would make sense.
> 
> The bundle size part of `origami-ci-tools` should probably be its own Github action.

yes! i agree

lib/helpers/files.js
```diff
@@ -44,87 +44,37 @@ function packageLockJsonExists(cwd) {
...
-					isInBowerMain = true;
-				} else if (typeof bowerJson.main === 'string' && bowerJson.main === 'main.js') {
-					isInBowerMain = true;
+	return getPackageJson(cwd)
+		.then(pkgJson => {
+			if (pkgJson) {
+				if (typeof pkgJson.main === 'string') {
+					return path.join(cwd, pkgJson.main);
+				} else if (typeof pkgJson.browser === 'string') {
+					return path.join(cwd, pkgJson.browser);
```
chee, line 66: i think we should prefer `browser` to `main`, because we're building for browsers

lib/tasks/demo-build.js
chee: maybe we should throw an error if a component depends on o-assets, explaining that o-assets is a tool for a bygone era and to talk to #origami-support about it

lib/helpers/files.js
JakeChampion: Yes, let's do that 👍

lib/tasks/demo-build.js
notlee: I don't think we need to add an error since components aren't using o-assets anyway, we should deprecate o-assets though!

notlee: https://github.com/Financial-Times/o-assets/pull/85

JakeChampion (commented):
Looks good to me 👍

MIGRATION.md
```diff
@@ -2,6 +2,14 @@
 
 ## Migrating from v10 to v11
 
+obt no longer supports bower, it now only supports npm.
```
JakeChampion, line 5: We should mention it only supports origami v2 and not origami v1 in this section

JakeChampion (commented):
One comment about the migration guide

MIGRATION.md
notlee: I figure we can rewrite the whole migration guide to be more clear when it's done :) I'm seeing it as a list of changes for future reference at the moment

PR #926, opened 2020-10-30
 Only lint files if the component itself contains linting config for those file types

JakeChampion:
This changes obt to only lint if the linting configs exist within the component project:
- .eslintrc.js to lint js
- .stylelintrc.js to lint sass/scss
- .remarkrc.js to lint markdown files like the readme

those linting tools do allow other config files such as .json instead of .js but I have not added support for that, should we support all the different options?


notlee:
Yikes. The tests are failing now because I stated using `npx stylelint`, which is no good for the unit tests. 
See commit for reason why: https://github.com/Financial-Times/origami-build-tools/pull/926/commits/476c832a6d3b3e84b739278ed063cfd23b84e651

>Stylelint plugins require `stylelint` which means if we’re using
>component lint config and obt doesn’t depend on the same plugins
>the plugin isn't resolved and an error is thrown
>
>To avoid this we can run the stylelint cli with `npx`. This
>feels not great but has some advantages. If a component depends
>on and configures a newer version of Stylelint obt won't error.

Maybe instead of this PR we should update Origami spec v2
to say Origami components must set an npm lint script. And
have build tools defer to that? Or we could remove linting
from build tools and have npm script defer to obt for other
features (like running demos)

chee:
> Yikes. The tests are failing now because I stated using `npx stylelint`, which is no good for the unit tests.
> See commit for reason why: [476c832](https://github.com/Financial-Times/origami-build-tools/commit/476c832a6d3b3e84b739278ed063cfd23b84e651)
> 
> > Stylelint plugins require `stylelint` which means if we’re using
> > component lint config and obt doesn’t depend on the same plugins
> > the plugin isn't resolved and an error is thrown
> > To avoid this we can run the stylelint cli with `npx`. This
> > feels not great but has some advantages. If a component depends
> > on and configures a newer version of Stylelint obt won't error.
> 
> Maybe instead of this PR we should update Origami spec v2
> to say Origami components must set an npm lint script. And
> have build tools defer to that? Or we could remove linting
> from build tools and have npm script defer to obt for other
> features (like running demos)

let's do that.

let's leave this situation as it is for origami v1, and for origami v2 we will have `npm run lint` which is a shortcut for stylelint, eslint, readme lint and demo linter.

later we can look at the scripts as their own project, and work out how to share them between projects if we desire. `"lint": "origami-scripts lint"`

:)

notlee:
I've opened a PR to move the tests which confirm our stylelint config to [stylelint-config-origami-component](https://github.com/Financial-Times/stylelint-config-origami-component/pull/31). When we do the above work we can confirm remove those tests from obt (keeping tests which confirm obt calls stylelint and outputs the results in a way we expect)

notlee:
This is still a bit rubbish since the if a component uses a new major of stylelint the message formatting obt does could break. Instead we could look at writting a formatter for stylelint instead and including it as a component dependency. But, I think our original aim is met here, and it's ready for review :)

test/integration/verify/verify.test.js
```diff
@@ -36,7 +36,8 @@ describe('obt verify', function () {
 					.then(obt => {
 						return execa(obt, ['verify']);
 					})
-					.then(() => {
+					.then((s) => {
+						console.log(s)
```
JakeChampion, line 40:
```suggestion
```

test/integration/verify/verify.test.js
```diff
@@ -209,7 +210,8 @@ describe('obt verify', function () {
 					.then(obt => {
 						return execa(obt, ['verify']);
 					})
-					.then(() => {
+					.then((s) => {
+						console.log({s})
```
JakeChampion, line 214:
```suggestion
```

test/integration/verify/verify.test.js
```diff
@@ -396,7 +398,8 @@ describe('obt verify', function () {
...
```
JakeChampion, line 402:
```suggestion
```

test/integration/verify/verify.test.js
```diff
@@ -36,7 +36,7 @@ describe('obt verify', function () {
 					.then(obt => {
 						return execa(obt, ['verify']);
 					})
-					.then(() => {
+					.then((s) => {
```
JakeChampion, line 39:
```suggestion
					.then(() => {
```

test/integration/verify/verify.test.js
```diff
@@ -209,7 +209,7 @@ describe('obt verify', function () {
...
```
JakeChampion, line 212:
```suggestion
					.then(() => {
```

test/integration/verify/verify.test.js
```diff
@@ -396,7 +396,7 @@ describe('obt verify', function () {
...
```
JakeChampion, line 399:
```suggestion
					.then(() => {
```

JakeChampion (commented):
I approve the PR but I cannot "approve" the PR because I raised it

notlee (approved):

Issue #876, opened 2020-08-10
Require component lint configuration


notlee:
If a component has an `.eslintrc.js` or `.stylelintrc.js` file that configuration is used for linting rules by the `obt verify` command. If not provided default configuration is used. This does/should [apply to `.remarkrc.js` too](https://github.com/Financial-Times/origami-build-tools/issues/875).

The team decided instead of falling back to default configuration origami-build-tools should throw an error. We think this is more transparent and ensures components provide local configuration, which can be used by other tools such as text editors to present errors to enginers before running origami build tools.

notlee:
Opps, Jake had started this but I picked it up again! 🙈 See: https://github.com/Financial-Times/origami-build-tools/pull/926

JakeChampion:
I believe we are now not going to add this feature and instead will add to the origami v2 spec that components should/must have their own `lint` npm run-script. -- https://github.com/Financial-Times/origami-build-tools/pull/926#issuecomment-731081612

notlee:
Let's keep this open until we create new issue(s). We should also capture in an issue that I've opened a PR to move the tests which confirm our stylelint config to stylelint-config-origami-component, they can be removed from obt:
https://github.com/Financial-Times/stylelint-config-origami-component/pull/31

notlee:
This work has been completed and will be released in the next major version (v11):
https://github.com/Financial-Times/origami-build-tools/pull/935

PR #930, opened 2020-11-18
Require component lint configuration.

notlee:
No longer provide default sass, js, or readme lint configuration
when not provided by the component.

We think this is more transparent and ensures components provide
local configuration, which can be used by other tools such as
text editors to present errors to engineers before running Origami
Build Tools.

However:
- The component spec does not require these files, therefore Origami
Build Tools technically would be unable to verify Origami compatible
components. I think this is technically bad but not actually bad.
- The error isn't super helpful. Since the verify tasks run in parallel
I don't think we can ask "should we add the missing config for you"?
Maybe a link somewhere or a dump of what config we expect could be
more helpful?

Note: This is a major release. Not all of our components have these
lint files. E.g. I notice `o-message` is missing the readme lint
config `.remarkrc.js`.

Closes: https://github.com/Financial-Times/origami-build-tools/issues/876

### DIFF
--- a/lib/tasks/verify-javascript.js
+++ b/lib/tasks/verify-javascript.js
@@ -18,22 +18,19 @@ async function lint (config) {
 		const hasGitignore = await fs.exists(gitignorePath);
 		const ignorePatterns = hasGitignore ? gitignore(await fs.readFile(gitignorePath)) : ['bower_components/', 'node_modules/'];
 
-		// Get eslint config from the component directory,
-		// or use the default if it does not exist.
-		const eslintPath = path.join(config.cwd, '/.eslintrc.js');
+		// Get eslint config from the component directory.
+		const eslintFile = '.eslintrc.js';
+		const eslintPath = path.join(config.cwd, eslintFile);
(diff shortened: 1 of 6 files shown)


lib/tasks/verify-javascript.js
```diff
@@ -18,22 +18,19 @@ async function lint (config) {
...
+		}
 
 		const eslint = new ESLint({
 			baseConfig: {
-				extends: ["origami-component"],
-				env: {
-					'browser': true,
-					'mocha': true
-				},
 				ignorePatterns,
```
notlee, line 31: this no longer includes bower by default, I think I can make a PR to our eslint conf for this

notlee: um no I'm not going there. We also want to ignore everything in .gitignore and I don't think that's possible through .eslintrc alone (there is a cli option tough)

notlee:
ugh Jake had already started this! https://github.com/Financial-Times/origami-build-tools/pull/926

PR #931, opened 2020-11-18
Update stylelint/eslint ignore paths.

notlee:
Use the component's `.gitignore` instead of parsing `.gitignore`
paths within origami build tools. This means `.eslintignore` is
not used and there is no fallback if `.gitignore` is not specified.

The latter is not a problem, the former maybe? The only component
which currently uses `.eslintignore` is `o-tracking`, and those
are old/`.gitignore` duplicates:
https://github.com/Financial-Times/o-tracking/blob/04862efac02f7fc952c505b9335a92b5166c0b34/.eslintignore#L2

### DIFF
--- a/lib/tasks/verify-javascript.js
+++ b/lib/tasks/verify-javascript.js
@@ -3,7 +3,6 @@
 const process = require('process');
 const path = require('path');
 const ESLint = require('eslint').ESLint;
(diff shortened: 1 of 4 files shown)


notlee:
see https://github.com/Financial-Times/origami-build-tools/pull/926

PR #942, opened 2021-01-11
Verify the component follows v2 of the Origami component specification.

notlee:
Rather than v1, which will not be supported in the next major version
of Origami Build Tools


JakeChampion (approved):
Nice work :-)

I think we could add some extra custom verify rules to help our components authors:
- Warn if a bower.json file exists and explain how to migrate to a package.json file
- Verify that the entry points in the package.json point to files which exist

notlee:
I made an issue for your suggestions 👉 https://github.com/Financial-Times/origami-build-tools/issues/943

PR #950, opened 2021-01-21
Add ability to publish prereleases with a different dist-tag.

JakeChampion:
Prior to this change, every release would be given the dist-tag `latest`, which is not what we want. We only want stable releases to be given the `latest` dist-tag.

I've changed the trigger for the workflows to no longer be on a created release but instead when a tag is pushed, this is because you can not filter on a release trigger but you can filter on a pushed tag.

Here is the documentation for filters --> https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

### DIFF
--- a/.github/workflows/publish-to-npm-as-latest.yml
+++ b/.github/workflows/publish-to-npm-as-latest.yml
@@ -1,9 +1,10 @@
-name: Publish to npm
+name: Publish to npm as latest version
 on:
-  release:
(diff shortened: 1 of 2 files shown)


notlee (approved):
this is excellent

origamiserviceuser:
:tada: This PR is included in version v10.8.9 :tada:

PR #952, opened 2021-01-21
Make git pushes in create-new-version workflow trigger other workflows

JakeChampion:
Currently we are using the read-only github token. When using this token and pushing commits/tags to github, github will not trigger any workflows for the pushed commit/tag.

We need the pushed tag to trigger out npm-publishing workflows, which means we need to use a custom github token instead of the default token provided by github actions.


notlee (approved):
oh ok (thanks for the link Jake https://docs.github.com/en/actions/reference/events-that-trigger-workflows#triggering-new-workflows-using-a-personal-access-token)

origamiserviceuser:
:tada: This PR is included in version v10.8.10 :tada:

PR #958, opened 2021-01-26
Replace Scrumple with esbuild and swc with babel

notlee:
Replace Scrumple with esbuild: Scrumple threw an error when
compiling a v2 spec component with lodash dependency. Instead
of figure that out we can switch to esbuild and reduce the amount
of custom tooling we need to maintain.

Replace swc with babel: This one is born of frustration. I found
it hard to change our implementation because documentation is a
work in progress. The sourcemaps produced by swc, using esbuild
sourcemaps as input, could not be interpreted by browsers. Babel
is slower but is more mature and in this case works without an
issue (in Chrome and Safari, Firefox still takes issue with
the sourcemap but is able to display the line of the error)


JakeChampion (approved):

origamiserviceuser:
:tada: This PR is included in version v11.0.0-beta.6 :tada:

PR #960, opened 2021-01-28
Run pa11y tests when there is no `pa11y.mustache` template.

notlee:
pa11y tests should run if there is a demo with name `pa11y` which
compiles to `pa11y.html` in the demos directory. The demo may or
may not be based on a template `pa11y.mustache`.

This commit also removes references to "custom config" (it would
have been awkward to commit separately). Custom config is a way
of configuring components outside of origami.json. This is not
documented anywhere in spec v1, moving to spec v2 is a good time
to drop suppport.

### DIFF
--- a/lib/helpers/files.js
+++ b/lib/helpers/files.js
@@ -91,6 +91,32 @@ function getModuleBrands(cwd) {
 		});
 }
 
+/**
+ * Get individual demo configuration for the component.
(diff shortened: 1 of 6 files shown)


JakeChampion (approved):

origamiserviceuser:
:tada: This PR is included in version v11.0.0-beta.8 :tada:

PR #961, opened 2021-02-05
inline the demo css and js into the html

JakeChampion:
This makes the demos each become a single file, which makes it possible for the origami build service to serve the single file via the /v3/demos endpoint (which is still being built).

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,7 @@ Origami Build Tools provides boilerplate for creating up a new Origami component
 		--browserstack             Run tests using Browserstack instead of Chrome Stable
 		--demo-filter=<demo-name>  Build a specific demo. E.G. --demo-filter=pa11y to build only the pa11y.html demo.
 		--brand=<brand-name>       Build SCSS for a given brand. E.G. --brand=internal to build the component for the internal brand.
+		--single-file-demo         Build the demo html as a single file which has the demo css and js inlined.
 		--debug                    Keep the test runner open to enable debugging in any browser.
 
 ## Commands
--- a/lib/helpers/files.js
+++ b/lib/helpers/files.js
@@ -44,8 +44,7 @@ function packageLockJsonExists(cwd) {
 	return fileExists(path.join(cwd || process.cwd(), '/package-lock.json'));
 }
 
-function getOrigamiJson(cwd) {
-	cwd = cwd || process.cwd();
+function getOrigamiJson(cwd = process.cwd()) {
 	return requireIfExists(path.join(cwd, '/origami.json'));
 }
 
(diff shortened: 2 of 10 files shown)


notlee:
Sorry I'm not sure I follow, could you expand on the reason for inlining? Just for my understanding, is this to match what will happen in the Build Service for the sake of consistency or a necessary change? And if so will v3 still error if there's a dependency conflict between component dependencies and demo dependencies as v2 currently does (I think)?

JakeChampion:
> Sorry I'm not sure I follow, could you expand on the reason for inlining? Just for my understanding, is this to match what will happen in the Build Service for the sake of consistency or a necessary change? And if so will v3 still error if there's a dependency conflict between component dependencies and demo dependencies as v2 currently does (I think)?

The way demos currently work on the origami build service is they use an old version of origami-build-tools which generates a html page which includes a set of script and link tags for the demo javascript and css similar to this:

```html
	<link rel="stylesheet" href="//www.ft.com/__origami/service/build/v2/bundles/css?modules=o-layout@4.2.0:/demos/src/demo.scss,o-table@^8.0.0,etc...">
	<script src="//www.ft.com/__origami/service/build/v2/bundles/js?modules=o-layout@4.2.0:/demos/src/demo.js,o-table@^8.0.0,etc..."></script>
```

Those urls use a feature of the build service where you can define which file within a package you want to include in the bundle, overriding the default entry point for the package. This is a feature I was wanting to avoid having in the new version of the origami build service as it's only reason to exist was to make demos work.

An alternative way to make demos work is to have the origami-build-tools have an option to inline the demo specific css and js into the html file as inline script and style tags instead of external script and link tags. This would enable the origami build service to be able to drop support for the feature which lets someone override the file a package should use as it's entry point when being bundled.

notlee:
Gotcha, I don't think obt does that locally, it includes two `link` tags. One to a local copy of the component demo and one to the Build Service for demo dependencies. E.g:
```html
<script src="https://polyfill.io/v3/polyfill.min.js?features=CustomEvent&flags=gated&unknown=polyfill"></script>
<script src="demo.js" defer></script>
<script src="https://www.ft.com/__origami/service/build/v2/bundles/js?modules=o-grid,o-colors" defer></script>
<link rel="stylesheet" href="https://www.ft.com/__origami/service/build/v2/bundles/css?modules=o-grid,o-colors" />
<link rel="stylesheet" href="demo.css" />
```
I'm happy to approve the PR and inline demo assets but now the Build Service is independent of Build Tools I don't think there's a need to?

JakeChampion:
I rewrote this to instead be a flag which can be provided.

Below is the commit message which explains my reasoning behind adding this flag.

>Add a new --single-file-demo flag for the demo task which will inline the demo css and js into the html.
>
>This makes the demos each become a single file, which makes it possible for the origami build service to serve the single file via the /v3/demos endpoint and not have to implement an api into the bundle endpoints which allows the request to decide which file in any component to include inthe bundle instead of the default entry-point file for the component.
>
>The Origami Build Service can then implement a demo endpoint which has this implementation:
>- Download component code from npm whose demo has been requested (this can be done via `npm pack package@version`
>- Unpack the component code into a folder
>- Run npm install within the folder
>- Run `obt demo --single-file-demo --demo-filter=<name of demo being requested> --brand=<brand being requested>`
>- Serve the html file that the previous command has built


@notlee I just saw your comment:
>I'm happy to approve the PR and inline demo assets but now the Build Service is independent of Build Tools I don't think there's a need to?

I forgot about trying to keep build tools separate from build service. I can duplicate the demo code from build tools into build service for sure, it would be pretty much a copy paste 👍.

notlee:
>I forgot about trying to keep build tools separate from build service. I can duplicate the demo code from build tools into build service for sure, it would be pretty much a copy paste 👍.

does this mean we can close this PR @JakeChampion?

PR #964, opened 2021-02-09
ensure npm 7 is used when running obt install

JakeChampion:
obt install is set up to use a locally installed version of npm if one exists. A local version of npm did not exist within the project and so obt install would fallback to using whatever `npm` would resolve to from the users PATH. This can cause issues for users if their global npm command is an old version of npm for example (This happened to me today).

By installing npm as a dependency to obt, we ensure that obt install will always use the version of npm that we want, which is npm version 7.


notlee (approved):

origamiserviceuser:
:tada: This PR is included in version v11.0.0-beta.10 :tada:

PR #966, opened 2021-02-09
make obt install always create a package-lock.json file

JakeChampion:
Having package-lock files for our components makes it much more likely that running `obt install` will end with a set of installed dependencies that are known to work.


chee (approved):

origamiserviceuser:
:tada: This PR is included in version v11.0.0-beta.12 :tada:

PR #969, opened 2021-02-11
prove that the pa11y demo does not need to be named pa11y.mustache

JakeChampion:
we don't define anywhere that the name of the demo file needs to be pa11y.mustache but we don't have a test which proves that the pa11y demo can have a different filename and work correctly.


chee (commented):
good work

PR #970, opened 2021-02-11
Make pa11y work on components which have no explicilty defined their brand support

JakeChampion:
we don't actually specify what the default brand is [if no brand is defined](https://origami.ft.com/spec/v1/manifest/#brands)...

We know it should be `master` though, so that is what this pull-request changes `getModuleBrands` to return if no brands are defined.

This also fixes a bug in the pa11y task where it would only run pa11y on the first brand defined for the component and not for all the brands defined.

### DIFF
--- a/lib/helpers/files.js
+++ b/lib/helpers/files.js
@@ -84,17 +84,18 @@ function getComponentName(cwd) {
 function getModuleBrands(cwd) {
 	return getOrigamiJson(cwd)
 		.then(origamiJson => {
-			if (origamiJson && origamiJson.brands && Array.isArray(origamiJson.brands)) {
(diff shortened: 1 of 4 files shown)


JakeChampion:
> We know it should be `master` though, so that is what this pull-request changes `getModuleBrands` to return if no brands are defined.

This is not true... if no brand is defined, it means it works with all brands.

JakeChampion:
The origami spec has been updated to reflect what having no `brands` property means:
https://origami.ft.com/spec/v1/manifest/#brands
screenshot of the brands section on https://origami.ft.com/spec/v1/manifest/#brands.

This is ready for review now 👍

chee (approved):

origamiserviceuser:
:tada: This PR is included in version v11.0.0-beta.13 :tada:

PR #978, opened 2021-02-23
pin all @swc/core subpackages to 1.2.47

JakeChampion:
We need to either pin these in package.json or use an npm-shrinkwrap.json because `npm install origami-build-tools` does not use package-lock.json.

The current version fails to run on the official nodejs docker images -- https://github.com/swc-project/swc/issues/1430


chee (approved):

origamiserviceuser:
:tada: This PR is included in version v10.8.13 :tada:

PR #984, opened 2021-03-16
expect origamiVersion to be the "2.0" string

chee:


lib/helpers/verify-component-specification-version.js
```diff
@@ -8,7 +8,7 @@ const { getOrigamiJson } = require('./files');
  */
 module.exports = async () => {
 	const origamiJson = await getOrigamiJson();
-	if (!origamiJson || typeof origamiJson.origamiVersion !== 'number') {
+	if (!origamiJson || typeof origamiJson.origamiVersion !== 'string') {
```
notlee, line 11: There is a condition after this which throws a more helpful error message when run obt is run on a v1 component, which won't be hit anymore

lib/tasks/verify-origami-json.js
```diff
@@ -26,8 +26,8 @@ function origamiJson(config) {
 						if (origamiJson.origamiType !== 'imageset' && origamiJson.origamiType !== 'module' && origamiJson.origamiType !== 'service') {
 							result.push('The origamiType property needs to be set to either "imageset", "module" or "service"');
 						}
-						if (origamiJson.origamiVersion !== 2) {
-							result.push('The origamiVersion property needs to be set to 2, this version of Origami Build tools only supports v2 of the Origami component specification.');
+						if (typeof origamiJson.origamiVersion !== 'string' || origamiJson.origamiVersion.split('.')[0] !== '2') {
+							result.push('The origamiVersion property needs to be set to "2.0" or higher, this version of Origami Build tools only supports v2 of the Origami component specification.');
```
notlee, line 30: non-blocking suggestion: might be helpful to mention that it needs to be a string here? Or maybe we should add a more specific error for when 2.0 is set as a Number?

chee: Added a message that says it must be a string. It will print both errors if they provide a number, so they will see that it must be the string and the format the string must be in.

lib/helpers/verify-component-specification-version.js
chee: thanks, i transposed them

lib/helpers/verify-component-specification-version.js
```diff
@@ -8,16 +8,6 @@ const { getOrigamiJson } = require('./files');
...
-		throw new Error(
-			'Is this an Origami component? \n\n' +
-			'Origami Build Tools may only be used to develop Origami ' +
-			'components. It looks like this isn\'t an Origami component, ' +
-			'or the component is missing a valid "origami.json". See the ' +
-			'component specification for more details: ' +
-			'https://origami.ft.com/spec/'
-		);
-	}
 	if (origamiJson.origamiVersion === 1) {
```
notlee, line 11: I think `getOrigamiJson` can return a null value 😬 if I remember right 

```suggestion
	if (origamiJson && origamiJson.origamiVersion === 1) {
```

notlee (approved):
🎉

PR #987, opened 2021-03-29
fix bug in karma config due to misuing the karma api

JakeChampion:
Newer versions of Karma throw an error if a file is supplied as a string instead of a file object.

Here we have updated our config to use the documented file object to ensure the newer versions of karma work with obt.

This is currently an issue everyone will hit if they are installing origami-build-tools for the first time because newer versions of karma will get installed (because we do not include an npm-shrinkwrap).

### DIFF
--- a/lib/tasks/karma.js
+++ b/lib/tasks/karma.js
@@ -38,7 +38,13 @@ const karmaTest = function (config, task) {
 
 			const { reporter, errors } = require('../plugins/listr-karma')();
 
-			karmaConfig.files.unshift(polyfillUrl);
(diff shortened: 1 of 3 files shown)


notlee (approved):

JakeChampion:
Note to self, we need to do this in major-v11 branch as well. We also need to use a shrinkwrap file.

JakeChampion:
Looks like this was added to major-v11 today via a merge commit -- https://github.com/Financial-Times/origami-build-tools/pull/935/commits/72f5655ea1a7dc2b9cb1d1a65918eec459d29905

PR #988, opened 2021-03-29
Swap to npm-shrinkwrap to ensure all installs use the same dependency versions

JakeChampion:
Previously we were using package-lock.json which ensures when running `npm install` in the origami-build-tools repo that it always installs the same dependency versions.

package-lock.json does not ensure that `npm install origami-build-tools` always uses the same dependency versions though, and that means when someone/something runs `npm install origami-build-tools` it will be using dependency versions we have not tested against and could be cause errors/bugs in our project. This has happened several times recently with Karma.

Using npm-shrinkwrap ensures that `npm install` and `npm install origami-build-tools` will both use the same dependency versions, which is what we want.


notlee (approved):
Nice. I'm still surprised by this

origamiserviceuser:
:tada: This PR is included in version v10.8.15 :tada:

PR #995, opened 2021-04-07
Improve tests for verify-package-json

JakeChampion:
For the scenarios where we were testing an invalid package.json file, if verify-package-json incorrectly validated the package.json, the test would not fail.

I've refactored those tests to now check if verify-package-json has returned a rejected promise, and if it has not, to fail.

I've also refactored the code to use proclaim.deepStrictEqual instead of proclaim.calledWithExactly because deepStrictEqual will the expected value and actual value if they are not the same, whereas calledWithExactly only states they are not the same, which I did not find useful at all because I couldn't tell where/how they were different.

### DIFF
--- a/test/unit/tasks/verify-package-json.test.js
+++ b/test/unit/tasks/verify-package-json.test.js
@@ -98,29 +98,29 @@ describe('verify-package-json', function () {
 			fs.writeFileSync('package.json', JSON.stringify({}), 'utf8');
 
 			let errored;
-				try {
-					await verifyPackageJson().task();
-					errored = false;
-				} catch (error) {
-					errored = true;
-					proclaim.equal(
-						error.message,
-						'Failed linting:\n\n' +
+			try {
+				await verifyPackageJson().task();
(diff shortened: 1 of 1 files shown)


test/unit/tasks/verify-package-json.test.js
```diff
@@ -82,8 +82,8 @@ describe('verify-package-json', function () {
 	describe('task', function () {
 		it('should run package.json check successfully', function () {
 			return verifyPackageJson().task().
-				then(function (verifiedOrigamiJson) {
-					proclaim.equal(verifiedOrigamiJson.length, 0);
+				then(function (verifiedpackageJson) {
+					proclaim.equal(verifiedpackageJson.length, 0);
 				});
```
chee, line 87:
```suggestion
				then(function (verifiedPackageJson) {
					proclaim.equal(verifiedPackageJson.length, 0);
				});
```

chee (approved):
better

PR #999, opened 2021-04-12
Make verify-origami-json look for all the different origamiType variations

JakeChampion:
When migrating ftdomdelegate we found that `obt verify` reports a failure because ftdomdelegate is listed as a library.

image

The three options I can think of to resolve this are:
1.  Make obt check for all the possible allowed values of origamiType
2.  Remove origamiType from the specification
3.  Remove obt from the projects which are not components, such as ftdomdelegate


chee (approved):

notlee (approved):

origamiserviceuser:
:tada: This PR is included in version v11.0.0-beta.21 :tada:

PR #1001, opened 2021-04-13
Only check packages are in the @financial-times scope if they are origami components

JakeChampion:
ftdomdelegate is an example of a library we maintain which uses obt to run it's tests. It is not an origami component nor is it under the @financial-times npm scope

The definition of origami component I'm using for this pull-request is `origami.json.origamiType` set to either `"module"` or `"component"`

Here is a list of the projects which are origami libraries and listed in the origami registry:
- ft-date-format
- get-github-public-organisation-repositories
- karma-proclaim
- node-health-check
- origami-navigation-data
- origami-repo-data-client-node
- origami-service
- polyfill-library
- sass
- streamcat
- svg-tint-stream

After looking through how all those projects are set up, none of them use obt, which makes me think we should not use obt for ftdomdelegate either

Should ftdomdelegate and the other origami libraries use origami-build-tools?

### DIFF
--- a/test/unit/tasks/verify-package-json.test.js
+++ b/test/unit/tasks/verify-package-json.test.js
@@ -19,7 +19,7 @@ const oTestPath = 'test/unit/fixtures/o-test';
 const pathSuffix = '-verify';
 const verifyTestPath = path.resolve(obtPath, oTestPath + pathSuffix);
 
-describe('verify-package-json', function () {
+describe.only('verify-package-json', function () {
 	let verifyPackageJson;
 	const originalConsole = global.console;
 	let console;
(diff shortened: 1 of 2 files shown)


test/unit/tasks/verify-package-json.test.js
```diff
@@ -19,7 +19,7 @@ const oTestPath = 'test/unit/fixtures/o-test';
 const pathSuffix = '-verify';
 const verifyTestPath = path.resolve(obtPath, oTestPath + pathSuffix);
 
-describe('verify-package-json', function () {
+describe.only('verify-package-json', function () {
```
JakeChampion, line 22:
```suggestion
describe('verify-package-json', function () {
```

JakeChampion:
We discussed on [Slack](https://financialtimes.slack.com/archives/CSW6B2VAN/p1618828191093400) and decided that non origami components should not use origami-build-tools.
That means this pull-request is not needed because those projects wouldn't use origami-build-tools in the first place 👍

PR #1004, opened 2021-04-21
Verify supported npm version is set to a minimum of 7.0.0

JakeChampion:
This verification helps ensure origami components adhere to the specification

The origami 2.0 spec states:
>It *must* include an engines property set to an object which has an npm property set to a SemVer range which disallows anything below v7.0.0. E.G. ^7 would be valid as would >= 7.


notlee (approved):

origamiserviceuser:
:tada: This PR is included in version v11.0.0-beta.24 :tada:

PR #1006, opened 2021-04-22
Allow './main.js' to be the browser entrypoint for a component

JakeChampion:
Currently we are checking it is exactly `'main.js'`, which means `'./main.js` would fail even though it points to the same file.

The solution I have provided might be too complicated, should the solution be to check if the field is either exactly `"main.js"` or `"./main.js"`?


notlee (approved):

test/unit/tasks/verify-package-json.test.js
```diff
@@ -535,6 +535,15 @@ describe('verify-package-json', function () {
...
 				});
+
+				it('should pass if property is set to `"./main.js"', async function () {
+					const packageJSON = JSON.parse(fs.readFileSync(path.join(process.cwd(), 'package.json'), 'utf-8'));
+					packageJSON.browser = './main.js';
+					fs.writeFileSync('package.json', JSON.stringify(packageJSON), 'utf8');
+
+					await verifyPackageJson().task();
+					proclaim.notCalled(console.log);
+				});
```
JakeChampion, line 546:
```suggestion
				it('should pass if property is set to `"./main.js"', async function () {
					const packageJSON = JSON.parse(fs.readFileSync(path.join(process.cwd(), 'package.json'), 'utf-8'));
					packageJSON.browser = './main.js';
					fs.writeFileSync('package.json', JSON.stringify(packageJSON), 'utf8');

					await verifyPackageJson().task();
					proclaim.notCalled(console.log);
				});

				it('should pass if property is set to an odd path which resolves to `"./main.js"', async function () {
					const packageJSON = JSON.parse(fs.readFileSync(path.join(process.cwd(), 'package.json'), 'utf-8'));
					packageJSON.browser = './src/../main.js';
					fs.writeFileSync('package.json', JSON.stringify(packageJSON), 'utf8');

					await verifyPackageJson().task();
					proclaim.notCalled(console.log);
				});
```

origamiserviceuser:
:tada: This PR is included in version v11.0.0-beta.26 :tada:

PR #1007, opened 2021-04-23
update all dependencies to latest versions which support commonjs

JakeChampion:
the aliases and strip-ansi packages have new versions which only support ecmascript modules, which means we can't update to those until we make origami-build-tools' code to use import/export


notlee (approved):

origamiserviceuser:
:tada: This PR is included in version v11.0.0-beta.25 :tada:

PR #1008, opened 2021-04-28
make demos use origami build service v3 for their dependencies instead of v2

JakeChampion:
Not sure how I missed this before

When migrating the components to origami 2.0, github ci showed test failures and percy failures which are because the demos were not using the correct version of the build service when including their dependencies

This pull-request fixes the issue 👍


origamiserviceuser:
:tada: This PR is included in version v11.0.0-beta.27 :tada:

PR #1013, opened 2021-05-06
Remove silent mode

notlee:
Remove silent mode for the primary mixin.

Origami spec v2 components will require a primary mixin and
will no longer recommend the silent mode variable.

Related proposal: https://github.com/Financial-Times/origami/pull/121
Related spec suggestion: https://github.com/Financial-Times/origami-website/pull/273#discussion_r626746297

_See commit messages for more re. the npm install change_

PR #1015, opened 2021-05-07
Use `require` over `import` in True Sass tests.

notlee:
I'm not sure when this broke. I'm not sure why it ever worked:
https://mochajs.org/#nodejs-native-esm-support


JakeChampion:
@notlee This shouldn't work for our components because they have `"type": "module"` defined in their `package.json`, which is why this file was changed to use `import` instead of `require` -- could you post a link or example of where this is not working? I think it works in our draft origami 2.0 components

JakeChampion:
> @JakeChampion thanks for the info. It's not working on the [obt integration tests here](https://github.com/Financial-Times/origami-build-tools/pull/1013). It's also not working for me locally when I try to run the test command against o-colors (the v2 branch) `CI=true /local/path/origami-build-tools/lib/origami-build-tools-cli.js test`. I'm running `v12.10.0` and `6.10.3` on this machine

could you try with npm 7 instead of npm 6? I'm not sure it will make a difference for this situation but origami 2.0 does not support npm 6

notlee:
Now with workflow updates 🙈

PR #1020, opened 2021-05-12
Update integration test temporary path. 

notlee:
OBT previously had three methods of creating temporary directories
to run tests in. Now each test uses the same helper. Temporary
directories are created within the obt integration test directory
which avoids cluttering our home directory and sidesteps any issues
with the temporary directory `os.tmpdir` we were seeing with Github
Actions.


JakeChampion (approved):
Nice tidy up 👍

PR #1022, opened 2021-05-13
Fix test task for IE11 / viewport dependent tests

notlee:
- Downgrade Sinon, sinon@10 drops IE11 support (there is a PR to
correct the site, which still says IE11 is supported).
- Bundle test JavaScript in the same way as the Build Service v3,
targeting es5. Previously es6+ syntax caused errors in IE11.
- Correct sourcemaps. Previously sourcemap related data was incorrectly
added to the page as text, pushing test DOM down, causing errors for
some component tests that depend on scroll position.
- Bundle and serve source javascript but do not include it. This
ensures when source code changes test code is processed again. I'm
unsure why this is required on top of watching the files, there may
be more efficient config to trigger test rebuilds on source file
changes.

Before: sinon errors in IE11, plus test bundle syntax errors, sourcemap junk text in the DOM
Screenshot 2021-05-13 at 16 45 36

Now: passing tests
Screenshot 2021-05-13 at 16 40 45


notlee:
We should follow up by sharing bundle code between the test/demo commands

lib/plugins/bundle-javascript.js
```diff
@@ -0,0 +1,58 @@
...
+			contents: compiledBundle.code
+		}
+	});
+
+	return Buffer.from(bundlefinal.outputFiles[0].contents).toString('utf-8');
+}
+
+module.exports = {
+	'preprocessor:javascript': ['factory', function bundleJavaScriptFactory() {
+		return async function hello(content, file, done) {
```
JakeChampion, line 53: :wave

JakeChampion (commented):
nice work figuring this out!

JakeChampion:
@notlee I added a beta release label to this

JakeChampion (approved):

lib/plugins/bundle-javascript.js
notlee: haha, oops

notlee:
```suggestion
		return async function (content, file, done) {
```

origamiserviceuser:
:tada: This PR is included in version v11.0.0-beta.32 :tada:

PR #1023, opened 2021-05-20
load the demo's dependency javascript before the demo's own javascript to ensure the dependecies are available to be used

JakeChampion:
Previously the demo javascript would load in before the dependencies javascript.

By swapping the order around, the demo javascript can now make use any variables the dependency javascript creates, such as `window.Origami` which contains all the dependencies' JS exports.


notlee (approved):

origamiserviceuser:
:tada: This PR is included in version v11.0.0-beta.33 :tada:

PR #1028, opened 2021-06-14
Update test command to only expect primary Sass mixins for components.

notlee:
Libraries may not have primary mixins. We would still expect them
to only output CSS when their Sass API is called though, not by
default on import.


origamiserviceuser:
:tada: This PR is included in version v11.0.0-beta.35 :tada:

PR #1031, opened 2021-06-17
Remove requirement that the primary mixin produces css by default

JakeChampion:
Not all components provide css by default for a particular brand, for example o-fonts' whitelabel mode has no css.
Percy's visual testing catches any regression a component would have if it used to produce css and no longer does


chee (approved):

origamiserviceuser:
:tada: This PR is included in version v11.0.2 :tada:

PR #1032, opened 2021-06-17
do not run sass compilation test if there is no main sass file

JakeChampion:
currently there is a bug in the test-sass-compilation module where it would try and run sass tests on libraries which have no sass such as o-autoinit and o-errors. This fixes the bug 👍


JakeChampion:
I'll add a test for this, it depends on making a version of o-test-component which is a library and has no sass -- https://github.com/Financial-Times/o-test-component/pull/183

notlee (approved):
nice 👍 

There's also the `fs/promises` api which I think is a bit easier than using denodeify :)

origamiserviceuser:
:tada: This PR is included in version v11.0.3 :tada:

PR #933, opened 2020-11-30
Make pa11y run against all supported brands and build the pa11y demo before it runs

JakeChampion:
This allows component which are built for a single brand to have a pa11y demo which works.

Solves - https://github.com/Financial-Times/origami-ci-tools/issues/23 and https://github.com/Financial-Times/o-meter/issues/2

This changes the pa11y task to not require `obt demo` to have been run first by the user.

PR #1034, opened 2021-07-07
Skip pa11y tests when the `--ignore-bower` flag is set.

notlee:
`origami-ci-tools` runs `obt test --ignore-bower`, to test the
npm package of a component which supports Bower and npm is correct:
https://github.com/Financial-Times/origami-ci-tools/blob/master/commands/branch.js#L52

This used to not run pa11y tests as the pa11y demo is not built by
`origami-ci-tools`, and Origami Build Tools used to skip pa11y tests
if a built pa11y demo was not found. However origami-build-tools
was updated to check for pa11y source code and build the pa11y
demo if it exists:
https://github.com/Financial-Times/origami-build-tools/pull/933/files#diff-6c44fc3cf122c89eee3ab192bd97d1a853d25adb56fa750332e79aac7ad230d4R90

`origami-component-converter`, used to generate the npm bundle, does
not support demo source code so the pa11y build fails, and thus
`origami-ci-tools` fails, and thus we can't make minor/patch release
to old versions of components which still support Bower.

By skipping pa11y tests if `--ignore-bower` is set we avoid this
error.

### DIFF
--- a/lib/tasks/pa11y.js
+++ b/lib/tasks/pa11y.js
@@ -81,19 +81,20 @@ module.exports = function (cfg) {
 					await buildDemo({
 						brand,
 						demoFilter: 'pa11y',
-						ignoreBower: config.ignoreBower,
 					});
 					return pa11yTest(config, brand);
 				}
 			}
 		},
-		skip: function () {
-			return fileExists(path.join(config.cwd, '/demos/src/pa11y.mustache'))
-				.then(exists => {
-					if (!exists) {
-						return `No Pa11y demo found. To run Pa11y against this project, create a file at ${path.join(config.cwd, '/demos/local/pa11y.html')}`;
-					}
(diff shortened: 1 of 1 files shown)


chee:
should we update origami-ci to stop doing `occ` stuff at all? we probably don't want these backports publishing to npm

chee (approved):

notlee:
> should we update origami-ci to stop doing `occ` stuff at all? we probably don't want these backports publishing to npm

I don't think it does any harm to but we probably don't need them publishing to npm, unless we ever need to backport for users including components via npm the old way. If we have any more issues with the deploy step I'll look at updating the workflows to not publish to npm instead :)

origamiserviceuser:
:tada: This PR is included in version v10.9.1 :tada:

PR #1035, opened 2021-07-14
close the filehandler that fileExists creates

JakeChampion:
NodeJS is deprecating the automatic closing of fileHandlers during garbage collection.

This commit updates the codebase to explicitly close the fileHandlers instead of relying on the deprecated behaviour.

Below is the message that is displayed when running obt prior to this commit:

>DeprecationWarning: Closing a FileHandle object on garbage collection is deprecated. 
>Please close FileHandle objects explicitly using FileHandle.prototype.close().
>In the future, an error will be thrown if a file descriptor is closed during garbage collection.

### DIFF
--- a/test/integration/helpers/fileExists.js
+++ b/test/integration/helpers/fileExists.js
@@ -1,16 +1,20 @@
 'use strict';
 
-const open = require('fs/promises').open;
+const { access } = require('fs/promises');
+const { constants } = require('fs');
 
 /**
  * Node.JS no longer has an fs.exists method.
- * Instead we use the fs.open method in read (`r`) mode.
- * fs.open will throw an error if the file does not exist.
+ * Instead we use the fs.access method and check we can read the file.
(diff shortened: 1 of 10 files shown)


test/integration/helpers/fileExists.js
```diff
@@ -11,6 +11,9 @@ const open = require('fs/promises').open;
 */
 module.exports = function fileExists(file) {
```
chee, line 12: We could use `fs.access` here instead, which doesn't open the file https://nodejs.org/api/fs.html#fs_fspromises_access_path_mode

chee (approved):

test/integration/helpers/fileExists.js
JakeChampion: Nice suggestion 🙂 I've updated the pull-request to use fs.access instead 👍

origamiserviceuser:
:tada: This PR is included in version v11.0.5 :tada:

Issue #529, opened 2018-05-18
demo data error


gvonkoss:
When there is a syntax error in e.g. demo.json, the following error appears:

```
 → Demo data not found: /path/to/demo.json
```

instead of information about the syntax error

chee:
this appears to still be a problem, here: https://github.com/Financial-Times/origami-build-tools/blob/3715752ecdc2a302a8d6803785a004f75c122bdd/lib/tasks/demo-build.js#L76-L86

PR #1038, opened 2021-08-16
Only run the rejection handler if the promise from reading the file was rejected

JakeChampion:
This resolves https://github.com/Financial-Times/origami-build-tools/issues/529

Previously the rejection handler was added via `.catch` whcih meant that any rejection whichs happened in the promise chain would get caught by the handler, this led to the other rejections being masked/overwritten by the handler.

Using the second argument to `.then` means the rejection handler will only catch rejections promise this particular promise and none of the others in the chain.


JakeChampion:
This needs to have a test written to show it fixes the issue and to help ensure we don't regress

notlee (approved):

origamiserviceuser:
:tada: This PR is included in version v11.0.7 :tada:

PR #1041, opened 2021-09-01
add only the root node_modules to the sass path

JakeChampion:
all sass dependencies should be peer-dependencies and so no nesting should happen, meaning they will all be in the root directory


chee:
we're doing monkey patches now

